PR #5, opened 2023-11-15
feat(postgres): sim support

Chriscbr:
Closes https://github.com/winglang/wing/issues/4914


postgres/lib.test.w
```diff
@@ -2,11 +2,10 @@ bring expect;
...
-if util.env("WING_TARGET") == "tf-aws" {
-  let db = new l.Database(name: "test", pgVersion: 15);
+let db = new l.Database(name: "test", pgVersion: 15);
 
-  test "run a simple query" {
-    let result = db.query("SELECT 1 as one, 2 as two;");
-    expect.equal(result.at(0), {one: 1, two: 2});
-  }
+test "run a simple query" {
+  util.sleep(1s);
```
eladb, line 8: Why is this needed?

postgres/lib.w
```diff
@@ -45,16 +42,52 @@ pub class Database {
   }
 }
 
-pub class DatabaseNeon impl IDatabase {
+class DatabaseSim impl IDatabase {
```
eladb, line 45: Maybe put under `db.sim.w`

postgres/lib.w
```diff
@@ -45,16 +42,52 @@ pub class Database {
...
+    let container = new containers.Workload(
+      name: "postgres",
+      image: image,
+      env: {
+        POSTGRES_PASSWORD: "password"
+      },
+      port: 5432,
+      public: true,
+    );
+    this.url = container.getPublicUrl();
```
eladb, line 58:
```suggestion
    this.url = container.publicUrl;
```

postgres/lib.w
```diff
@@ -45,16 +42,52 @@ pub class Database {
...
+}
+
+struct DbCredentials {
+  host: str;
+  user: str;
+  password: str;
+  database: str;
+}
+
+class DatabaseNeon impl IDatabase {
```
eladb, line 81: `db.neon.w'

eladb (approved):
Super nice

postgres/lib.test.w
Chriscbr: Ah I think I was debugging something, it should be possible to remove it

postgres/lib.w
Chriscbr: I was curious what's the difference between them

eladb: It's because Interfaces don't support fields and I wanted to enforce the api.

Pretty ugly.

postgres/lib.test.w
Chriscbr: Looks like we do need some kind of command-based readiness check.. for now I'm just retrying connecting

postgres/lib.w
eladb: Yes, in other projects we needed to put the interface into a separate file.

I am okay to leave as is

Chriscbr: Props would have to go in a separate file as well

eladb: crap

PR #6, opened 2023-11-18
feat(redis): sim support

subh-cs:
Redis as a third-party Wing library.

`redis` has all the available methods from ex.Redis(). 

Related to https://github.com/winglang/wing/issues/2171


wing-redis/README.md
```diff
@@ -0,0 +1,53 @@
+# Wing Redis Client Library
+
+This library provides a Wing client for Redis.
+
+## Installation
+
+Use `npm` to install this library:
+
+```sh
+npm i @winglibs/wing-redis
```
eladb, line 10: Please rename to just `redis`

wing-redis/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Bring it
+
+The `redis.Redis` resource represents a Redis client.
+
+```js
+bring wing-redis;
+
+new redis.Redis() as "redis";
```
eladb, line 20:
```suggestion
new redis.Redis();
```

Maybe also add a simple example of usage in this initial code block.

wing-redis/README.md
```diff
@@ -0,0 +1,53 @@
...
+* `del` - Deletes a value from the Redis server.
+* `hget` - Gets a value from a hash on the Redis server.
+* `hset` - Sets a value on a hash on the Redis server.
+* `sadd` - Adds a value to a set on the Redis server.
+* `smembers` - Gets all members of a set on the Redis server.
+
+
+## `sim`
+
+When executed in the Wing Simulator, the client is started within a local Docker container.
```
eladb, line 38:
```suggestion
When executed in the Wing Simulator, a Redis server is started within a local Docker container.
```

wing-redis/example.main.w
```diff
@@ -0,0 +1,52 @@
...
+new cloud.Function(inflight () => {
+    redis1.set(key, value);
+    log("value set to redis1 successfully");
+}) as "setValToRedis1";
+
+new cloud.Function(inflight () => {
+    let value = redis1.get(key);
+    log("value get from redis1 ${value}");
+}) as "getValFromRedis1";
+
```
eladb, line 20: Covert to a .test.w file with tests

wing-redis/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@winglibs/wing-redis",
+  "version": "0.0.1",
+  "description": "Redis client for Wing",
```
eladb, line 4:
```suggestion
  "description": "Redis resource for Wing",
```

wing-redis/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@winglibs/wing-redis",
```
eladb, line 2:
```suggestion
  "name": "@winglibs/redis",
```

wing-redis/redis.sim.w
```diff
@@ -0,0 +1,88 @@
...
+    redisImage: str;
+    state: sim.State;
+
+    init() {
+    this.redisContainerName = "wing-sim-redis-${this.node.id}-${util.uuidv4()}";
+    // Redis version 7.0.9
+    this.redisImage = "redis@sha256:e50c7e23f79ae81351beacb20e004720d4bed657415e68c2b1a2b5557c075ce0";
+    this.state = new sim.State();
+    // start and stop the instance
+    let s = new cloud.Service(inflight () => {
```
eladb, line 18: Please use `containers.Workload` instead of hand rolling. 

See the Postgres library as an example.

wing-redis/redis.test.w
```diff
@@ -0,0 +1,20 @@
+bring expect;
+bring "./redis.w" as wing_redis;
```
eladb, line 2:
```suggestion
bring "./redis.w" as r;
```

This is becoming our convention for now

wing-redis/redis.test.w
```diff
@@ -0,0 +1,20 @@
...
+let redis = new wing_redis.Redis();
+
+test "set and get" {
+  redis.set("foo", "bar");
+  expect.equal(redis.get("foo"), "bar");
+}
+
+test "set, get and del" {
+  redis.set("foo", "bar");
+  redis.del("foo");
```
eladb, line 13: Add an expectation that the value exists before it's deleted

wing-redis/utils.w
```diff
@@ -0,0 +1,14 @@
+bring "./api.w" as api;
+bring fs;
+
+pub class Util {
+  extern "./utils.js" pub static inflight shell(command: str, args: Array<str>, cwd: str?): str;
```
eladb, line 5: This will not be needed if you are using containers.Workload

wing-redis/utils.js
```diff
@@ -0,0 +1,54 @@
...
+            if (error) {
+                console.error(stderr);
+                return reject(error);
+            }
+
+            return resolve(stdout ? stdout : stderr);
+        });
+    });
+};
+
```
eladb, line 17: I would go with a different approach:

```js
interface IRedisClient {
  inflight get(key: str): str;
  inflight set(key: str, ...): void;
  // ....
}

class Util {
  extern "./redis.js"
  static inflight newClient(url): IRedisClient;
}
```

And then, in redis.js:

```js
const IoRedis = require("ioredis");

exports.newClient = (url) => new IoRedis(url);
```

And that's it. Basically then you can just use the client from wing.

eladb (changes requested):

wing-redis/README.md
subh-cs: I have changed it manually everywhere possible. Should I change the `wing-redis.yaml` file manually at `.github/workflows`?

redis/README.md
```diff
@@ -0,0 +1,57 @@
+# Wing Redis Client Library
```
eladb, line 1:
```suggestion
# Redis support for Winglang
```

redis/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `redis.Redis` resource represents a Redis client.
+
+```js
+bring redis;
+
+let redis  = new redis.Redis();
+
+new cloud.Function(inflight () => {
+    redis.set(key, value);
```
eladb, line 23:
```suggestion
  redis.set("mykey", "myvalue");
```

1. We use 2 spaces
2. The example should work

redis/redis.sim.w
```diff
@@ -0,0 +1,88 @@
+bring "./api.w" as api;
+bring "./utils.w" as utils;
+bring cloud;
+bring util;
+bring sim;
+
+pub class Redis_sim impl api.IRedis {
+    redisContainerName: str;
```
eladb, line 8: Change to 2 spaces

redis/redis.sim.w
```diff
@@ -0,0 +1,88 @@
...
+    redisImage: str;
+    state: sim.State;
+
+    init() {
+    this.redisContainerName = "wing-sim-redis-${this.node.id}-${util.uuidv4()}";
```
eladb, line 13: Indentation is corrupted

redis/redis.sim.w
```diff
@@ -0,0 +1,88 @@
...
+bring cloud;
+bring util;
+bring sim;
+
+pub class Redis_sim impl api.IRedis {
+    redisContainerName: str;
+    redisImage: str;
+    state: sim.State;
+
+    init() {
```
eladb, line 12:
```suggestion
    new() {
```

Please use the latest wing version

redis/redis.sim.w
```diff
@@ -0,0 +1,88 @@
...
+    inflight start(){
+        try{
+            // check if image exists
+            utils.shell("docker", ["image", "inspect", this.redisImage]);
+            log("image already exists ${this.redisImage}");
+        }catch{
+            // pull image
+            log("pulling image ${this.redisImage}");
+            utils.shell("docker", ["pull", this.redisImage]);
+        }
```
eladb, line 34: Please use containers.Workload

eladb (changes requested):
See unresolved comments from previous revision

redis/redis.test.w
```diff
@@ -0,0 +1,21 @@
+bring expect;
+bring "./redis.w" as r;
+
+let redis = new r.Redis();
```
subh-cs, line 4: In this line, the redis container has started and port is 32719,

redis/redis.test.w
```diff
@@ -0,0 +1,21 @@
...
+
+test "set and get" {
```
subh-cs, line 6: But, when I'm starting this test, it is removing the previous container of port 32719, and starting a new container with port 32720.

redis/redis.test.w
```diff
@@ -0,0 +1,21 @@
...
+  redis.set("foo", "bar");
+  expect.equal(redis.get("foo"), "bar");
+}
+
+test "set, get and del" {
```
subh-cs, line 11: The same thing happening here as well. But the IoRedis client is looking for the container with the port it's first initialized with which doesn't exist anymore, causing `[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:32775`

redis/redis.test.w
subh-cs, line 1: How can I fix this?

redis/redis.test.w
eladb: indeed, in wing tests are isolated. so every time a test is executed, the entire preflight environment is created, so you know you have a clean setup.

(think of preflight as the "setup" of each test in traditional testing frameworks)

redis/redis.sim.w
```diff
@@ -0,0 +1,55 @@
...
+bring cloud;
+bring util;
+bring sim;
+bring containers;
+
+pub class Redis_sim impl api.IRedis {
+
+    connectionUrl: str;
+
+    new(){
```
eladb, line 12: please use 2 spaces for indentation

redis/redis.sim.w
```diff
@@ -0,0 +1,55 @@
...
+        let image = "redis@sha256:e50c7e23f79ae81351beacb20e004720d4bed657415e68c2b1a2b5557c075ce0";
+        let container = new containers.Workload(
+            name: "redis",
+            image: image,
+            port: 6379,
+            public: true,
+        );
+        this.connectionUrl = container.publicUrl ?? "";
```
eladb, line 20:
```suggestion
        this.connectionUrl = container.publicUrl ?? "error";
```

until we have `?? throw "error"`

redis/redis.js
```diff
@@ -0,0 +1,6 @@
+const IoRedis = require("ioredis")
+
+exports.newClient = (url) => {
+    let port = url.split(":")[2];
```
eladb, line 4: 2 spaces

redis/redis.js
```diff
@@ -0,0 +1,6 @@
...
+    return new IoRedis(port)
+}
```
eladb, line 6:
```suggestion
};
```

redis/redis.js
```diff
@@ -0,0 +1,6 @@
+const IoRedis = require("ioredis")
+
+exports.newClient = (url) => {
+    let port = url.split(":")[2];
+    return new IoRedis(port)
```
eladb, line 5:
```suggestion
    return new IoRedis(port);
```

redis/redis.sim.w
eladb, line 12:
```suggestion
    new() {
```

redis/redis.test.w
eladb: > But the IoRedis client is looking for the container with the port it's first initialized

why is it using the previous port?

eladb (changes requested):

redis/redis.test.w
subh-cs: Hi, it's making a socket connection with the server, and when server is not there, it's keep looking for the server. I've figured out a way though, it's working but, 
```js
let port = url.split(":")[2];
  let redis = new IoRedis(port)
  redis.on("error", (err) => {  // by this, we can listen if there's any error
    console.log(`Redis error: ${err}`);
    redis.disconnect(); // this will disconnect the connection between server and the client, and it will stop looking for server. 
  });
  return redis;
```
But the problem is `redis.disconnect()` is disconnecting all the available clients. 

During wing test, wing is creating multiple redis container, therefore multiple clients has been created with diff port, and when one of the container is getting killed and hitting the "error" event to trigger `redis.disconnect()`, it is deleting all the client beforehand even they start executing the tests, so the error is coming `connection is closed` for next 2 tests (first tests passing successfully)

So, now I think the only way to somehow disconnect only one client at a time. But I'm not able to figure out how. 
somthing like this. 
```js 
redis.disconnect(port); // though it is not a valid 
```
Can you please let me know if you have any suggestions, idea on how can I make this work, though I've just joined official redis discord and put this query to their channel, hope they help.

eladb: I am not sure I fully understand what's going on here. If for every test we start a new container on a new port and there's a client that connects to that server, why are they all disconnected together?

Does this problem also exists in our current [`ex` implementation](https://github.com/winglang/wing/blob/main/libs/wingsdk/src/target-sim/redis.inflight.ts)? 

Copying @yoav-steinberg who wrote the `ex` implementation.

subh-cs: Hi, I have found the problem.

When I'm running the test at wing-console, it is passing successfully and at each test I'm running shell command `docker ps`, and you can see the log, the url port is same as the docker instance is running.

Screenshot from 2023-11-26 13-57-47

But the same test file when I'm running by the command from terminal `wing test redis.test.w`, as you can see the url port is different than the docker instance running, that specific docker instance is closed in this case.
image
 
I think this is a problem with tests, let me know what can be a solution of that. I was wrong about that the problem is redis client specific but it is not.

redis/api.w
```diff
@@ -0,0 +1,25 @@
...
+  inflight hget(key: str, field: str): str?;
+  inflight sadd(key: str, value: str): void;
+  inflight smembers(key: str): Array<str>?;
+  inflight del(key: str): void;
+}
+
+// TODO: remove this once we have a proper way to extend interfaces
+pub interface IRedisClient extends IRedis {
+  inflight disconnect(): void;
+  // remove them
```
eladb, line 15: ??

eladb (changes requested):
@subh-cs I don't think this fully closes  https://github.com/winglang/wing/issues/2171 because this only supports `sim` and the `ex` version also supports `tf-aws`.

Let's only close that issue when we've finalized.

redis/api.w
eladb: This bug was fixed today https://github.com/winglang/wing/pull/5074

eladb: @subh-cs can you confirm this has been resolved?

subh-cs: No, it is not resolved. Currently getting an different error message. checkout dev channel, I've posted it there.

subh-cs: Hi @eladb, it is a bug https://github.com/winglang/wing/issues/5089, though I know it is similar to this https://github.com/winglang/wing/issues/5052.

redis/redis.js
```diff
@@ -0,0 +1,6 @@
+const IoRedis = require("ioredis")
```
eladb, line 1:
```suggestion
const IoRedis = require("ioredis");
```

redis/redis.sim.w
```diff
@@ -0,0 +1,70 @@
...
+bring "./utils.w" as utils;
+bring cloud;
+bring util;
+bring sim;
+bring containers;
+
+pub class Redis_sim impl api.IRedis {
+  connectionUrl: str;
+
+  new(){
```
eladb, line 11:
```suggestion
  new() {
```

redis/redis.sim.w
```diff
@@ -0,0 +1,70 @@
...
+    let image = "redis:latest";
+    let container = new containers.Workload(
+      name: "redis",
+      image: image,
+      port: 6379,
+      public: true,
+    );
+    this.connectionUrl = container.publicUrl ?? "error";
+  }
+
```
eladb, line 21: you should be able to use an inflight constructor and inflight field to persist the client across calls:

```js
class Redis_sim {
  inflight client: IRedisClient;
  inflight new() {
    this.client = utils.newClient(this.connectionUrl);
  }

  pub inflight get(key: str): str? {
    return this.client.get(key);
  }

  // ...
}
```

eladb (changes requested):

eladb (approved):

PR #12, opened 2023-11-23
feat(containers): improve dx by skipping image pull

subh-cs:
fixes: https://github.com/winglang/wing/issues/5019

Whenever workload starts, docker will be checking every time before pulling an image from registry, providing better dx.


eladb (approved):
Nice

eladb (approved):

PR #16, opened 2023-11-25
feat: `bring vite`

skyrpex:
Allows using a [Vite](https://vitejs.dev/) project with Wing.


skyrpex:
View the [README.md](https://github.com/winglang/winglibs/blob/add-vite-resource/vite/README.md).

skyrpex:
Workarounds I had to do:

- In order to use the Vite JS API to start the dev server and build, I need to use  `spawn` and `execSync` (so esbuild bundling doesn't interfere, and so async functions become sync)
- In order to know the URL of the Vite dev server, I need to find a free port and pass it to Vite. Otherwise, there's no sane way for me to retrieve it from Wing's runtime context

I believe we would increase the DX a lot if Wing was async in preflight (by embracing ESM), and async during the constructors.

vite/vite.w
```diff
@@ -0,0 +1,139 @@
...
+  env: Map<str>;
+  cli: str;
+  generateTypeDefinitions: bool;
+}
+
+class SimVite {
+	pub url: str;
+
+	new(props: SimViteProps) {
+		let port = new find_port.Port();
```
eladb, line 18: something is off with indentation here...

vite/vite.w
```diff
@@ -0,0 +1,139 @@
...
+bring "./find-port.w" as find_port;
+
+struct SimViteProps {
+  root: str;
+  env: Map<str>;
+  cli: str;
+  generateTypeDefinitions: bool;
+}
+
+class SimVite {
```
eladb, line 14:
```suggestion
class Vite_sim {
```

(this is the weird convention we had for this)

vite/vite.w
```diff
@@ -0,0 +1,139 @@
+bring cloud;
+bring sim;
+bring util;
+bring fs;
+bring "./find-port.w" as find_port;
+
+struct SimViteProps {
+  root: str;
+  env: Map<str>;
```
eladb, line 9:
```suggestion
  env: Map<str>?;
```

vite/cli.mjs
```diff
@@ -0,0 +1,49 @@
+import { parseArgs } from "node:util";
```
eladb, line 1: doesn't vite have a cli? why do we need to wrap it?

vite/env-plugin.mjs
```diff
@@ -0,0 +1,88 @@
+import { mkdir, writeFile } from "node:fs/promises";
+
+const PACKAGE_NAME = "@wingcloud/vite";
```
eladb, line 3: ??

vite/env-plugin.mjs
```diff
@@ -0,0 +1,88 @@
+import { mkdir, writeFile } from "node:fs/promises";
```
eladb, line 1: what's this?

vite/vite.w
```diff
@@ -0,0 +1,139 @@
...
+
+    let target = util.env("WING_TARGET");
+    if target == "sim" {
+			let implementation = new SimVite(
+        root: props.root,
+        env: env,
+        cli: cli,
+        generateTypeDefinitions: generateTypeDefinitions,
+      );
+			this.url = implementation.url;
```
eladb, line 110: indentation

eladb (changes requested):
a few minor comments

vite/vite.w
skyrpex: I made it required for the target implementations because there's some env variables we are forced to pass, otherwise things won't work (eg, `PATH` and `HOME`). Both targets needed them.

eladb: in this case i would "elevate" them into explicit apis such as `path` and `home` and then plug them into the environment below.

vite/env-plugin.mjs
skyrpex: Oops, it should be `@winglibs/vite`.

vite/vite.w
skyrpex: Fair enough.

PR #17, opened 2023-11-26
chore: modernize for wing v0.50.4

eladb:
- replace `${}` with `{}`
- replace `init` with `new`
- some changes around `std.Node.of()`.

bump all libs that were changed


containers/workload.tfaws.w
```diff
@@ -68,7 +68,7 @@ pub class Workload_tfaws impl api.IWorkload {
         );
     
         deployment.addContainer(
-          image: "{{ .Values.image }}",
+          image: "\{\{ .Values.image }}",
```
staycoolcall911, line 71: Need to escape just the opening `{` and not the closing `}`?

staycoolcall911 (approved):

containers/workload.tfaws.w
eladb: yes ([playground](https://www.winglang.io/play/?code=dABlAHMAdAAgACIAZQBzAGMAYQBwAGUAIgAgAHsACgAgACAAbABvAGcAKAAiAGgAZQBsAGwAbwAgAFwAewB3AG8AcgBsAGQAfQAiACkAOwAKAH0A))

ShaiBer (approved):

PR #23, opened 2023-11-28
feat(websocket): add websocket library

marciocadev:
This pull request adds a WebSocket resource to Wing.

The library can be used on `sim`, `tf-aws`, and `awscdk` targets.

I've copied a significant portion of the implementation created by @eladcon for the `sim` target.

Thanks, Elad!


websocket/inflight/websocket.sim.mts
```diff
@@ -0,0 +1,16 @@
+import { WebSocketServer } from "ws";
+
+export const _sendMessage = (connectionId: string, message: string) => {
+  let wss: WebSocketServer = global.wss;
```
skyrpex, line 4: Is `global.wss` something that the `ws` package creates?

websocket/README.md
```diff
@@ -0,0 +1,73 @@
+## Wing WebSocket Support
+
+This library allows you tu create WebSocket usiing Wing.
+
+## Installation
+
+Use `npm` to iinstall this library:
```
skyrpex, line 7:
```suggestion
Use `npm` to install this library:
```

websocket/README.md
```diff
@@ -0,0 +1,73 @@
...
+});
+
+wb.onMessage(inflight (id: str, body: str): void => {
+  let connections = tb.scan();
+  for item in connections.items {
+    wb.sendMessage(str.fromJson(item.get("connectionId")), body);
+  }
+});
+
+wb.initialize();
```
skyrpex, line 54: Is this necessary?

websocket/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "directory": "websocket"
+  },
+  "author": {
+    "email": "marciocadev@gmail.com",
+    "name": "Marcio Cruz de Almeida"
+  },
+  "license": "MIT",
+  "peerDependencies": {
+    "cdktf": "^0.19.1",
+    "winglang": "^0.47.11"
```
skyrpex, line 18: Maybe it should be upgraded to latest wing version?

websocket/platform/sim.w
```diff
@@ -0,0 +1,62 @@
...
+    this.disconnectFn = handler;
+  }
+  pub onMessage(handler: inflight(str, str): void): void {
+    this.messageFn = handler;
+  }
+
+  pub initialize() {
+    new cloud.Service(inflight () => {
+      let res = WebSocket_sim._startWebSocketApi(this.connectFn, this.disconnectFn, this.messageFn);
+      this.bucket.put("url.txt", res.url());
```
skyrpex, line 36: I think we could use a `sim.State` here.

ekeren: I agree, here is a reference how it is being used 

https://github.com/winglang/winglibs/blob/main/containers/workload.sim.w

websocket/README.md
ekeren: I think it is required because we are not using `sim.state`, using the token mechanism of sim.State should make this API obselete

marciocadev: The issue is that `cloud.Service` requires `connectFn`, `disconnectFn`, and `messageFn` to be filled, but they are `inflight(str): void` variables

`sim.State.set` only accepts json.

 I attempted to parse them, but it resulted in errors.

marciocadev: The `initialize` is necessary because I need the inflight functions to be defined before initializing the `cloud.Service`. Since these functions are configured in the preflight, it's not possible to use `sim.State.set` for configuration

And `sim.State.set` accepts `json`, not `inflight(str): void`.

websocket/platform/sim.w
marciocadev: see above

websocket/inflight/websocket.sim.mts
marciocadev: No, `global.wss` is a variable created in the global scope that can be referenced elsewhere.

In this case, it was created during the initialization of the websocket in file `platform/sim/wb.mts`

https://github.com/marciocadev/winglibs/blob/d27705faf91fe16a90840e9eb48a774c45d6f25f/websocket/platform/sim/wb.mts#L9-L11

websocket/README.md
marciocadev: @eladcon open this issue about it

[captured field is undefined even when initialized](https://github.com/winglang/wing/issues/4324)

eladb: this shouldn't be needed

websocket/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+When executing in the Wing Simulator, the WebSocket uses the `ws` library.
+
+## `tf-aws`
+
+When running on AWS, the WebSocket utilizes the WebSocket API of API Gateway.
+
+## `awscdk`
+
+When running on AWS, the WebSocket utilizes the WebSocket API of API Gateway.
```
eladb, line 67: please add a "# Maintainers" section here (see example in the bedrock library)

websocket/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@winglibs/websocket",
+  "version": "0.0.1",
+  "description": "WebSocket support for Wing",
```
eladb, line 4:
```suggestion
  "description": "WebSocket library for Wing",
```

websocket/README.md
ekeren: @eladb ,  it is required today because of https://github.com/winglang/wing/issues/4324. 
Do you have a workaround?

websocket/platform/sim.w
eladb: I think @skyrpex means to use `sim.State` for the url (instead of the bucket)

websocket/platform/sim/wb.mts
```diff
@@ -0,0 +1,34 @@
+import { WebSocketServer } from "ws";
+
+export const _startWebSocketApi = (
+  onConnect: (connectionId: string) => void,
+  onDisconnect: (connectionId: string) => void,
+  onMessage: (connectionId: string, message: string) => void) => {
+
+  const port = Math.floor(Math.random() * 1000 + 3000);
```
eladb, line 8: what happens if the random port is already used? isn't it possible to get the server select a port automatically?

websocket/README.md
eladb: yes, you are right.

another approach is to allow people to hook into the synthesis phase (e.g. override something like `onSynth()` to perform these types of finalization activities.

i think we need to fix #4324 *and* expose `onSynth()` to allow for both approaches.

eladb: @marciocadev it is undocumented, but in the meantime, if you create a method called `_preSynthesize` it will be called just before synthesis.

[see playground](https://www.winglang.io/play/?code=YwBsAGEAcwBzACAARgBvAG8AIAB7AAoAIAAgAF8AcAByAGUAUwB5AG4AdABoAGUAcwBpAHoAZQAoACkAIAB7AAoAIAAgACAAIAB0AGgAcgBvAHcAIAAiAGgAZQBsAGwAbwAiADsACgAgACAAfQAKAH0ACgAKAG4AZQB3ACAARgBvAG8AKAApADsA)

i will create a PR to turn this into a public method that people can use.

eladb (changes requested):

websocket/README.md
marciocadev: @eladb, I changed the `pub initialize()` to `_preSynthesize()` and it started giving me the following error when I try to use the simulator in the test

As in my test, I use Dynamo to capture who is logged in, it is not able to resolve the table at the pre-synthesis moment

image

eladb: i see... okay - keep the `initialize()` for now and i'll take a look at a proper solution.

thanks!

websocket/commons/api.w
```diff
@@ -0,0 +1,20 @@
+pub struct RouteOptions {
+  routeKey: str;
+}
+
+pub interface IWebSocket extends std.IResource {
```
eladb, line 5: please add docstrings here (they are not respected yet but soon)

@Chriscbr

websocket/commons/api.w
```diff
@@ -0,0 +1,20 @@
...
+  onMessage(handler: inflight(str, str): void): void;
+
+  initialize(): void;
+
+  inflight sendMessage(connectionId: str, message: str);
+
+  inflight url(): str;
+}
+
+pub struct WebSocketProps {
```
eladb, line 17: is this part of the public api?

websocket/platform/awscdk.w
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  pub inflight url(): str {
+    return this.invokeUrl;    
+  }
+
+  extern "../inflight/websocket.aws.mts" static inflight _postToConnection(endpointUrl: str, connectionId: str, message: str): void;
+  pub inflight sendMessage(connectionId: str, message: str) {
+    WebSocket_awscdk._postToConnection(this.callbackUrl, connectionId, message);
+  }
```
eladb, line 165: @MarkMcCulloh this makes me wonder if perhaps we can change `extern` to be something like this:

```js
  pub inflight sendMessage(connectionId: str, message: str) {
    extern "../inflight/websocket.aws.mts" _postToConnection(endpointUrl: str, connectionId: str, message: str): void;
    
    _postToConnection(this.callbackUrl, connectionId, message);
  }
```

websocket/commons/api.w
marciocadev: Yes, it's the props to create the WebSocket. If I remove the `pub`, the classes that implement the interface cannot receive the `props` defined in this file.

websocket/README.md
```diff
@@ -0,0 +1,77 @@
+## Wing WebSocket Support
+
+This library allows you tu create WebSocket usiing Wing.
+
+## Installation
+
+Use `npm` to install this library:
+
+```sh
+npm i @winglibs/websocket
```
eladb, line 10:
```suggestion
npm i @winglibs/websockets
```

websocket/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+## Bring it
+
+The example above shows us how we can broadcast a message to any connection assigned to the WebSocket.
+
+```js
+bring cloud;
+bring ex;
```
eladb, line 19: Missing:

```suggestion
bring ex;
bring websockets;
```

websocket/README.md
```diff
@@ -0,0 +1,77 @@
+## Wing WebSocket Support
+
+This library allows you tu create WebSocket usiing Wing.
```
eladb, line 3: Add a little background and some links about websockets and why would someone wants to use it

websocket/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+When running on AWS, the WebSocket utilizes the WebSocket API of API Gateway.
+However, to run on `awscdk`, we need to import the `@winglang/platform-awscdk`.
+
+```bash
+$ npm i @winglang/platform-awscdk
+$ CDK_STACK_NAME=MySocketStack wing compile -t @winglang/platform-awscdk websocket.main.w
+```
+
+## Maintainers
```
eladb, line 75: Missing license

websocket/README.md
```diff
@@ -0,0 +1,77 @@
...
+});
+
+wb.onMessage(inflight (id: str, body: str): void => {
+  let connections = tb.scan();
+  for item in connections.items {
+    wb.sendMessage(str.fromJson(item.get("connectionId")), body);
+  }
+});
+
+wb.initialize();
```
eladb, line 54: Add a comment explaining that this is temporarily needed

websocket/README.md
```diff
@@ -0,0 +1,77 @@
...
+When executing in the Wing Simulator, the WebSocket uses the `ws` library.
+
+## `tf-aws`
+
+When running on AWS, the WebSocket utilizes the WebSocket API of API Gateway.
+
+## `awscdk`
+
+When running on AWS, the WebSocket utilizes the WebSocket API of API Gateway.
+However, to run on `awscdk`, we need to import the `@winglang/platform-awscdk`.
```
eladb, line 68:
```suggestion
To compile to `awscdk`, we need to import the `@winglang/platform-awscdk`.
```

websocket/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@winglibs/websocket",
```
eladb, line 2:
```suggestion
  "name": "@winglibs/websockets",
```

websocket/platform/aws/api.w
```diff
@@ -0,0 +1,21 @@
+bring cloud;
+bring "../../commons/api.w" as api;
+
+pub struct WebSocketAwsRequestContext {
```
eladb, line 4: Add docstrings everywhere

websocket/platform/sim.w
```diff
@@ -0,0 +1,61 @@
...
+    this.disconnectFn = handler;
+  }
+  pub onMessage(handler: inflight(str, str): void): void {
+    this.messageFn = handler;
+  }
+
+  pub initialize() {
+    new cloud.Service(inflight () => {
+      let res = WebSocket_sim._startWebSocketApi(this.connectFn, this.disconnectFn, this.messageFn);
+      this.state.set("invokeUrl", res.url());
```
eladb, line 37: Put `"invokeUrl"` in a const

PR #32, opened 2023-12-10
feat: `bring lock`

ekeren:


lock/README.md
```diff
@@ -0,0 +1,57 @@
...
+bring lock;
+bring cloud;
+bring util;
+
+let l = new lock.Lock();
+let q = new cloud.Queue();
+
+q.setConsumer(inflight (m: str) => {
+  let id = util.uuidv4();
+  l.acquire("my-lock", 20s);
```
Chriscbr, line 25: What happens if a function crashes while it is holding the lock? Does the hold on the lock have an automatic expiry of some kind? Without some automatic expiry or heartbeat, I'm not sure if this resource would be useful in production.

lock/lib.w
```diff
@@ -0,0 +1,74 @@
...
+        }
+        let preDec = this.counter.dec(1, id);
+        // randomize sleep time to prevent livelock
+        // util.sleep(duration.fromMilliseconds(100 * math.random()));
+        return false;
+      }, timeout: timeout);
+      if acquired {
+        return;
+      }
+      throw "Failed to acquire lock, timeout {timeout.seconds} seconds reached";
```
Chriscbr, line 47:
```suggestion
      if !acquired {
        throw "Failed to acquire lock, timeout {timeout.seconds} seconds reached";
      }
```

lock/README.md
```diff
@@ -0,0 +1,57 @@
...
+  log("{id}:start");
+  util.sleep(1s);
+  log("{id}:end");
+  l.release("my-lock");
```
Chriscbr, line 29: Does this mean anybody can release anyone else's lock while it is still being held? Is there a way we can prevent this?

ekeren: Interesting 🤔 

Do you think this is a better API? 
```
let myLock = l.lock("name");
myLock.unlock();
```

Chriscbr: I prefer the Java-style API, but it could be a familiarity thing. In my head, if a thread or process or inflight isn't obtaining an exclusive handle when it acquires a lock, then it's less of a "lock" and more of an atomic flag, or mutex of some sort.

I'm not all that sure though so I'm ok trying this out

Chriscbr (approved):

eladb:
@ekeren are you planning to merge this?

ekeren:
Keep , on my todo list

ekeren:
ouch, not getting to this... lets try again

PR #33, opened 2023-12-11
fix(websockets): change all mts files to js

marciocadev:
The `wing pack` command doesn't support `mts` files, so I need to change them to `js` files

and exposing the WebSocket url to allow a React project to receive the url through the environment


websockets/winglibs-websockets-0.0.2.tgz
Chriscbr, line 1: Please remove

eladb (approved):

PR #39, opened 2023-12-19
feat: `bring github`

eladcon:


github/README.md
```diff
@@ -0,0 +1,23 @@
+# github
+
```
eladb, line 2: Please add some context and description

github/README.md
```diff
@@ -0,0 +1,23 @@
...
+`sh
+npm i @winglibs/github
+`
+
+## Usage
+
+`js
+bring github;
+
+let adder = new github.Adder();
```
eladb, line 18: Please update with documentation

github/lib.test.w
```diff
@@ -0,0 +1,76 @@
+bring fs;
+bring expect;
+bring util;
+bring "./probot/probot.w" as probot;
+let key = fs.readFile("/path/to/private-key.pem");
+
+let handler = inflight (ctx) => {
+  fs.writeFile("/tmp/111", "pull requesdt opened");
```
eladb, line 8: Probably better to use mkdtemp to put this is a unique temp dir

github/lib.w
```diff
@@ -0,0 +1,7 @@
+bring "./probot/probot.w" as probot;
+
+pub class ProbotApp extends probot.ProbotApp {
```
eladb, line 3: I am wondering if this needs library needs to be called `probot` instead of GitHub.

eladb, line 3: What's the reason we have this class?

github/octokit/types.w
```diff
@@ -0,0 +1,311 @@
...
+	inflight getCommit(options: GetCommitProps): GetCommitResponse;
+	inflight compareCommits(options: CompareCommitsProps): CompareCommitsResponse;
+	inflight createOrUpdateFileContents(options: CreateOrUpdateFileContentsProps): BaseResponse;
+	inflight createForAuthenticatedUser(options: CreateRepoProps): BaseResponse;
+	inflight createInOrg(options: CreateOrgRepoProps): BaseResponse;
+	inflight delete(options: DeleteRepoProps): BaseResponse;
+	inflight listForAuthenticatedUser(options: ListForAuthenticatedUserProps): ListReposResponse;
+	inflight listForOrg(options: ListForOrgProps): ListReposResponse;
+}
+
```
eladb, line 304: Is this generated somehow or manually written?

github/probot/types.w
```diff
@@ -0,0 +1,86 @@
+bring "../octokit/types.w" as octokit;
+
+struct IProbotRepositoryOwner {
```
eladb, line 3: This is a data type, not an interface, so the I prefix is not needed

```suggestion
struct ProbotRepositoryOwner {
```

eladb: Also below...

eladb (changes requested):

github/lib.test.w
ekeren: It is a debug entry, we will remove it

github/lib.test.w
```diff
@@ -0,0 +1,76 @@
...
+
+  // find all changed mdfiles by comparing the commits of the PR
+  let compare = ctx.octokit.repos.compareCommits(
+    owner: repo.owner.login,
+    repo: repo.name,
+    base: ctx.payload.pull_request.base.sha,
+    head: ctx.payload.pull_request.head.sha,
+  );
+
+  fs.writeFile("/tmp/112", "{Json.stringify(compare.data.commits)}");
```
ekeren, line 19: Remove

github/lib.w
eladb: Okay.

I guess the readme will be useful so I can understand how this is supposed to be used.

github/lib.test.w
ekeren, line 8:
```suggestion
```

github/lib.test.w
ekeren, line 19:
```suggestion
```

ekeren (approved):

PR #40, opened 2023-12-19
chore: set up stale PR workflow

Chriscbr:
Set up a workflow for auto-closing stale pull requests

### DIFF
--- a/.github/workflows/close-stale.yaml
+++ b/.github/workflows/close-stale.yaml
@@ -1,0 +1,27 @@
+name: Close Stale PRs
+on:
+  schedule:
+    - cron: 0 6 * * *
+jobs:
+  stale:
+    name: Closing stale PRs
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/close-stale.yaml
```diff
@@ -0,0 +1,27 @@
...
+            Hi,
+
+
+            This PR has not seen activity in 20 days. Therefore, we are marking the PR as stale for now. It will be closed after 7 days.
+
+            If you need help with the PR, do not hesitate to reach out in the winglang community slack at [winglang.slack.com](https://winglang.slack.com).
+
+            Feel free to re-open this PR when it is still relevant and ready to be worked on again.
+
+            Thanks!
```
MarkMcCulloh, line 27: Is an extra newline being added for each line?

MarkMcCulloh (approved):

.github/workflows/close-stale.yaml
Chriscbr: I want to say this won't have extra newlines since value in the YAML file starts with `>-` instead of `>`,  but let's see what happens

PR #46, opened 2023-12-31
fix(websockets)!: fix token usage in simulator

eladb:
BREAKING CHANGE: the `url()` and `inflightUrl()` methods have been removed in favor of a `url` property.

### DIFF
--- a/websockets/lib.w
+++ b/websockets/lib.w
@@ -6,16 +6,24 @@ bring "./platform/sim.w" as sim;
 
 pub class WebSocket impl api.IWebSocket {
   inner: api.IWebSocket;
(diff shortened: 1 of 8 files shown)


websockets/lib.w
```diff
@@ -6,16 +6,24 @@ bring "./platform/sim.w" as sim;
 
 pub class WebSocket impl api.IWebSocket {
   inner: api.IWebSocket;
+
+  pub url: str;
```
staycoolcall911, line 10: Need to bump the version on package.json

staycoolcall911 (approved):

PR #60, opened 2024-01-16
fix(ngrok): reference `ngrok.js` as a relative path

eladb:


skyrpex:
Yup, tried using the lib but failed..

skyrpex:
I think this solution won't work since `__dirname` is used during inflight, which is a bundled function and can't retain bundle-less info like dirname or filename.

See https://github.com/winglang/winglibs/pull/79, it gathers `__dirname` during preflight as a solution.

eladb:
Yes. I saw. Great job!

PR #61, opened 2024-01-17
chore: update wing version

polamoros:
- Updates wing to the latest version in each winglib.
- Adds a script to update wing (would be nice to do it in a cron job).

### DIFF
--- a/update-wing.sh
+++ b/update-wing.sh
@@ -7,7 +7,7 @@ for dir in */ ; do
             echo "Updating winglang in '$dir'..."
             npm update winglang
         else
-            echo "winglang not found in '$dir'."
+            echo "winglang not found in '$dir'.\n"
         fi        
         cd -
     fi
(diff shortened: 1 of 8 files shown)


update-wing.sh
```diff
@@ -0,0 +1,14 @@
...
+for dir in */ ; do
+    if [ -f "$dir/package.json" ]; then
+        cd "$dir"        
+        if grep -q "\"winglang\"" package.json; then
+            echo "Updating winglang in '$dir' winglib..."
+            npm update winglang
+        else
+            echo "winglang not found in '$dir' winglib."
+        fi        
+        cd ..
```
skyrpex, line 12: `cd -` seems safer to me, in case we change the directory layout.

polamoros: yep 👍

skyrpex (approved):

PR #63, opened 2024-01-20
feat: `bring typescript` (WIP)

eladb:
WIP


typescript/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```ts
+// src/index.ts
+exports.handler = async (event, context) => {
+  console.log("Welcome to TypeScript", event);
+  return {
+    status: 200,
+    body: JSON.stringify({ message: "Hello, TypeScript" }),
+  };
+};
```
Chriscbr, line 30: Can we add types to this example?

As a user I'd want some kind of information or disclaimer about what the signature of the function should be and whether it's enforced by the Wing library or not

typescript/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+```js
+// main.w
+bring typescript;
+
+new cloud.Function(new typescript.FunctionHandler("./src"));
```
Chriscbr, line 37: (optional) I wonder if it's weird to recommend one lambda function per folder

```suggestion
new cloud.Function(new typescript.FunctionHandler("./src/index"));
```

eladb: yes!

ekeren: >  I wonder if it's weird to recommend one lambda function per folder

@Chriscbr, How would this work for API gateway?

eladb: done

typescript/example/index.ts
```diff
@@ -0,0 +1,20 @@
+import { Handler } from "./index.wing";
+
+export const handler: Handler = async ({ myBucket, myQueue }, event: any) => {
```
MarkMcCulloh, line 3: Other than the IDE, would this be type-checked at any point during compilation?

eladb: At the moment, no. We just bundle with esbuild. I guess we can add a typecheck pass. I'll add a TODO

typescript/example/index.wing.d.ts
```diff
@@ -0,0 +1 @@
+export type Handler = (context: any, event: any) => Promise<any>;
```
eladb, line 1: This will be generated with types based on lifting

PR #64, opened 2024-01-23
feat: expose request headers

polamoros:


eladb (approved):

polamoros:
Not needed.

We are sending a JWT (which includes the userId) in the first message after establishing the ws connection, so we can establish a relationship between the connectionId and the user.

we don't need to access the request headers for auth.

PR #69, opened 2024-02-05
feat: `bring sagemaker`

tsuf239:
fixes: https://github.com/winglang/wing/issues/5565
@ekeren FYI

### DIFF
--- a/sagemaker/README.md
+++ b/sagemaker/README.md
@@ -1,0 +1,28 @@
+# sagemaker
+
+## Prerequisites
+
+- [winglang](https://winglang.io).
+
+## Installation
+
+`sh
+npm i @winglibs/sagemaker
+`
+
+## Usage
+
+`js
+bring sagemaker;
+
+let sm = new sagemaker.Endpoint("my-endpoint-name", "my-inference-name);
+
+let handler = inflight () => {
+let res = sm.invoke({inputs: "do AI stuff"});
+log(res.Body);
+}
+`
+
+## License
+
+This library is licensed under the [MIT License](./LICENSE).
--- a/sagemaker/README.md
+++ b/sagemaker/README.md
@@ -12,7 +12,9 @@ npm i @winglibs/sagemaker
 
 ## Usage
 
-`js
+The library enables owners of a trained sagemaker model, to access its Endpoints from a winglang inflight code.
+
+```js
 bring sagemaker;
 
 let sm = new sagemaker.Endpoint("my-endpoint-name", "my-inference-name);
@@ -21,7 +23,7 @@ let handler = inflight () => {
 let res = sm.invoke({inputs: "do AI stuff"});
 log(res.Body);
 }
-`
+```
 
 ## License
 
--- a/sagemaker/sagemaker.test.w
+++ b/sagemaker/sagemaker.test.w
@@ -1,12 +1,21 @@
+bring cloud;
 bring expect;
 bring "./sagemaker.w" as sageMaker;
 bring util;
 
 let sm = new sageMaker.Endpoint(util.tryEnv("ENDPOINT_NAME") ?? "Example-endopint", util.tryEnv("INFERENCE_NAME") ?? "Example-inference");
 
+let invokeModel = inflight (input: str) => {
+  let var s = input;
+  if input.length == 0 {
+    s = "Hi There!";
+  }
+  let res = sm.invoke( { inputs: "{s}" }, { ContentType: "application/json" });
+  log(res.Body);
+};
 
-test "testing endpoint" {
-  let res = sm.invoke( { inputs: "Hello there!" }, { ContentType: "application/json" });
+new cloud.Function(invokeModel);
 
-  log(Json.stringify(res));
+test "main" {
+    invokeModel("Hello there!");
 }
--- a/sagemaker/sagemaker.test.w
+++ b/sagemaker/sagemaker.test.w
@@ -2,7 +2,7 @@ bring expect;
 bring "./sagemaker.w" as sageMaker;
 bring util;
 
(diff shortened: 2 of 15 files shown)


sagemaker/README.md
```diff
@@ -0,0 +1,28 @@
+# sagemaker
+
+## Prerequisites
+
+- [winglang](https://winglang.io).
+
+## Installation
+
+`sh
```
eladb, line 9: Fix the backticks

sagemaker/README.md
```diff
@@ -0,0 +1,28 @@
...
+npm i @winglibs/sagemaker
+`
+
+## Usage
```
eladb, line 13: Please expand the documentation and explain the various APIs

sagemaker/sagemaker.test.w
```diff
@@ -0,0 +1,12 @@
+bring expect;
+bring "./sagemaker.w" as sageMaker;
+bring util;
+
+let sm = new sageMaker.Endpoint(util.env("ENDPOINT_NAME"), util.env("INFERENCE_NAME"));
```
eladb, line 5: How would this test work without the environment variables set?

Ideally a test should be self contained.

sagemaker/sagemaker.tfaws.w
```diff
@@ -0,0 +1,32 @@
...
+    if let lambda = aws.Function.from(host) {
+      lambda.addPolicyStatements({
+        actions: ["sagemaker:InvokeEndpoint"],
+        effect: aws.Effect.ALLOW,
+        resources: [
+          "arn:aws:sagemaker:*:*:endpoint/{this.endpointName.lowercase()}",
+          "arn:aws:sagemaker:*:*:inference-component/{this.inferenceComponentName.lowercase()}"
+        ]
+      });
+    }
```
eladb, line 27: Throw an exception if this is the wrong target

sagemaker/sagemaker.w
```diff
@@ -0,0 +1,37 @@
...
+
+
+    let target = util.env("WING_TARGET");
+    if target == "sim" {
+      if std.Node.of(this).app.isTestEnvironment {
+        // in case of test running on sim, use simulator version
+        this.inner = new s.SageMaker_sim(endpointName, inferenceComponentName) as "sim"; 
+      } else {
+        // in case of running on sim interactively (in development mode), use AWS version
+        this.inner = new t.SageMaker_tfaws(endpointName, inferenceComponentName) as "tf-aws";
```
eladb, line 24:
```suggestion
        this.inner = new t.SageMaker_tfaws(util.env("SAGEMAKER_ENDPOINT_NAME"), inferenceComponentName) as "tf-aws";  
```

Maybe the right DX is to allow this to be defined through environment variables when developing interactively.

sagemaker/types.w
```diff
@@ -0,0 +1,22 @@
+pub struct invocationOptions {
```
eladb, line 1:
```suggestion
pub struct InvocationOptions {
```

sagemaker/types.w
```diff
@@ -0,0 +1,22 @@
...
+  CustomAttributes: str?;
+  TargetModel: str?;
+  TargetVariant: str?;
+  TargetContainerHostname: str?;
+  InferenceId: str?;
+  EnableExplanations: str?;
+  InferenceComponentName: str?;
+}
+
+pub struct invocationOutput {
```
eladb, line 13:
```suggestion
pub struct InvocationOutput {
```

eladb (changes requested):

sagemaker/sagemaker.tfaws.w
tsuf239: I think @ekeren wants it to be able to be used from the sim target too

sagemaker/sagemaker.w
eladb: Good point about multiple instances, but I think it's still useful to support somehow specifying these externally to the app because then users would need a special case for the simulator

tsuf239: I agree, but I think it's part of a bigger "problem" that is shared across a few libraries with a similar structure (which I copied from bedrock), I think I'll open an issue for that :)

tsuf239: https://github.com/winglang/wing/issues/5629

eladb: Maybe this is something we can support through `wing.yaml` that @hasanaburayyan is working on

hasanaburayyan: 👋 Howdy, yea https://github.com/winglang/wing/pull/5557 introduces some features that could potentially help out here. 

For instance there is a `parameterRegistrar` that exists on the root App, that can be used to register/read parameters. Though, its currently marked as internal (which maybe it shouldnt be)

so for instance you could register this parameter with something like: 
`std.Node.of(this).app.parameterRegistrar.addParameterSchema(some_json_schema)`

The validation of the parameters happens before the construct tree is synthesized so, we can tell users exactly what they are missing. Also potentially in another iteration add tools for providing these parameters interactively.

tsuf239: and we can also have a default value, right? I'll add another issue to change it here and on bedrock as soon as it is merged :)

tsuf239: right now we can workaround it by:
```
let var endpointName = util.env("ENDPOINT_NAME"); // prod one
if (!app.isTestEnvironment &&  util.env("WING_TARGET") == "sim") {
  endpointName = <some value read from some file>; // sim non-test one
}
```
if needed

hasanaburayyan: > and we can also have a default value

Technically JSON schema supports explicitly stating defaults, however they have no real effect. So even if the schema defined a default it would still be up to your code to enforce it. (its essentially just docs)

PR #72, opened 2024-02-09
fix: update fifoqueue and github to latest version of Wing

Chriscbr:


Chriscbr:
Smells like a compiler bug, investigating

Chriscbr:
Depends on changes in https://github.com/winglang/wing/pull/5683

eladb (approved):

PR #76, opened 2024-02-12
chore(sagemaker): added missing region for troubleshooting

ekeren:


tsuf239 (approved):

sagemaker/README.md
```diff
@@ -32,6 +32,40 @@ let handler = inflight () => {
...
+Encountering issues? Here are some common problems and their solutions:
+
+### `Error: Region is missing`
+
+**Problem**: When trying to invoke the sagemaker model, the region reports missing
+`Error: Region is missing`
+
+**Solution**:
+
+Add `AWS_REGION` to the cloud.Function environment variable:
```
eladb, line 46: I don't think that's the "right" solution. You don't want to hard code the region in your code.

This is the region for development, not necessarily production.

I think you should just add the region to your environment when running `wing` or is it not prorated into the inflight?

ekeren: > You don't want to hard code the region in your code.

💯 

It is not propogated, any suggestion beside doing? 
`"AWS_REGION" : util.env("AWS_REGION")`

tsuf239: I think the sagemaker client support region: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sagemaker-runtime/
do we have it on the app instance somewhere? (we won't have it on the sim app for sure)

PR #77, opened 2024-02-13
feat: `bring messagefanout`

marciocadev:
This library connects a topic (SNS) directly to one or more queues (SQS), functioning as a 'message-fanout'.

To add a consumer (SQS) to this component, the preflight method is used:
`addConsumer(handler: inflight(str): void, props: MessageFanoutProps): void;`

To send a message, the inflight method is used:
`inflight publish(message: str): void;`

### This implementation supports the `tf-aws` and `sim` targets


message-fanout/platform/sim.w
```diff
@@ -0,0 +1,13 @@
...
+bring "../commons/api.w" as api;
+
+pub class MessageFanout_sim impl api.IMessageFanout {
+  topic: cloud.Topic;
+
+  new() {
+    this.topic = new cloud.Topic();
+  }
+
+  pub addConsumer(handler: inflight(str): void, props: api.MessageFanoutProps): void {}
```
hasanaburayyan, line 11: So this does not do anything in the sim currently?

message-fanout/README.md
```diff
@@ -0,0 +1,23 @@
...
+npm i @winglibs/message-fanout
+`
+
+## Usage
+
+`js
+bring message-fanout;
+
+let adder = new message-fanout.Adder();
+`
```
hasanaburayyan, line 19: can you add some examples of this adding consumers

message-fanout/platform/sim.w
marciocadev: Sorry Hasan, I had forgotten to implement it, now it's correct

message-fanout/README.md
marciocadev: Now it has a better example, Hasan.

hasanaburayyan (approved):
@marciocadev Thanks for fixing up the sim impl 🔥 LGTM

PR #78, opened 2024-02-14
fix: sagemaker and checks breakages

Chriscbr:
Updating code in `sagemaker` and `checks` winglibs to work with the latest version of Wing.

### DIFF
--- a/checks/check.w
+++ b/checks/check.w
@@ -88,7 +88,7 @@ pub class Check impl ICheck {
    * Invokes the check and returns it's result
    */
   pub inflight run(): r.CheckResult {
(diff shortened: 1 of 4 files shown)


checks/check.w
```diff
@@ -88,7 +88,7 @@ pub class Check impl ICheck {
    * Invokes the check and returns it's result
    */
   pub inflight run(): r.CheckResult {
-    let result = this.invoker.invoke("");
+    let result = this.invoker.invoke("")!;
```
eladb, line 91:
```suggestion
    let result = this.invoker.invoke()!;
```

PR #82, opened 2024-02-16
feat: `bring vite`

skyrpex:


.github/workflows/vite-pull.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          node-version: 20.x
+          registry-url: https://registry.npmjs.org
+      - name: Install winglang
+        run: npm i -g winglang
+      - name: Install dependencies
+        run: npm install
+        working-directory: vite
+      - name: Test
+        run: wing test **/*.test.w
```
MarkMcCulloh, line 25: This isn't needed, `wing test` will grab all tests in the directory for you

```suggestion
        run: wing test
```

skyrpex: Thanks! I didn't think of changing the script too much.

vite/vite.cjs
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  return new Promise((resolve) => {
+    child.stdout.on("data", (chunk) => {
+      const payload = parsePayload(chunk.toString());
+      if (payload) {
+        resolve(payload.url);
+        child.stdout.removeAllListeners("data");
+      } else {
+        // console.log(chunk.toString());
```
ainvoner, line 49: do we need it?

ainvoner (approved):

vite/vite.cjs
skyrpex: Not really, although I think we should have a way to log whatever Vite is outputting somewhere. I had an issue in Wing to discuss this, somewhere. I'll look for it.

PR #84, opened 2024-02-18
chore: disable container canary tests

Chriscbr:
These tests are failing intermittently - I couldn't seem to figure out the root cause so this will disable them for now.

Opened a separate issue to track this: https://github.com/winglang/wing/issues/5716

Misc:
- Updated package lock files

### DIFF
--- a/.github/workflows/bedrock-pull.yaml
+++ b/.github/workflows/bedrock-pull.yaml
@@ -3,8 +3,6 @@ on:
   pull_request:
     paths:
       - bedrock/**
-    paths-ignore:
-      - bedrock/package-lock.json
 jobs:
   build-bedrock:
     runs-on: ubuntu-latest
--- a/.github/workflows/bedrock-pull.yaml
+++ b/.github/workflows/bedrock-pull.yaml
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - bedrock/**
+    paths-ignore:
+      - bedrock/package-lock.json
 jobs:
   build-bedrock:
     runs-on: ubuntu-latest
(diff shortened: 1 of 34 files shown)


.github/workflows/bedrock-pull.yaml
```diff
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - bedrock/**
+    paths-ignore:
+      - bedrock/package-lock.json
```
MarkMcCulloh, line 7: why the ignore? if the lockfile changed should the pr check to see that the new versions in the lockfile work?

MarkMcCulloh (approved):

.github/workflows/bedrock-pull.yaml
Chriscbr: My motivation was that a change to the lockfile shouldn't result in a new attempted release of a library (since it doesn't get included in the tarball). But I guess it's okay if it still results in a PR build - I'll change it

MarkMcCulloh: Even though a lockfile doesn't get included in the tgz, a change in a lockfile can still result in a change to the final build artifact. Both transitive deps and loosely pinned deps can change without a change to the package.json.

Chriscbr: Huh, really? That's interesting/surprising to me. I imagined that would only apply if the dependencies were bundled?

MarkMcCulloh: It's for the build/test tools rather than the runtime dependencies. It's like the node, npm, or linux version used to build the package. If those change it can technically change the resulting artifact. or different transitive testing dependencies can cause tests to fail.

PR #98, opened 2024-02-23
feat: `bring dynamodb`

skyrpex:
- Interface similar to the original DynamoDB SDK
- Includes streams support
- Uses one container for all tables (sim)

Closes https://github.com/winglang/wing/issues/4949


ekeren:
@skyrpex what is the status of this?

skyrpex:
Hmmm not much I think! 

- Extract interface & abstract implementation from current table class
- Implement tf-aws table
- Marshall dynamodb stream event
- Not sure if we're missing critical type defnitions or methods

ekeren:
Sounds promising, does it also work locally?

skyrpex:
> Sounds promising, does it also work locally?

Yes, it works locally :)

dynamodb/dynamodb.mjs
```diff
@@ -0,0 +1,152 @@
...
+      shardIterator = recordsData.NextShardIterator;
+
+      // AWS DynamoDB fetches records 4 times a second.
+      await new Promise((resolve) => setTimeout(resolve, 250));
+    }
+  }
+};
+
+const processRecords = async (endpoint, tableName, handler) => {
+  while (true) {
```
eladb, line 109: How do we break this loop on shutdown?

dynamodb/dynamodb.mjs
```diff
@@ -0,0 +1,152 @@
...
+  }
+};
+
+export const processRecordsAsync = async (endpoint, tableName, handler) => {
+  processRecords(endpoint, tableName, handler).catch((error) => {
+    if (error.message.includes("ECONNREFUSED")) {
+      // Ignore. This error happens when reloading the console.
+      // We can safely end the execution here.
+    } else {
+      console.error(error);
```
eladb, line 149: No need to throw?

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+bring util;
+bring cloud;
+bring "./dynamodb-types.w" as dynamodb_types;
+bring "./dynamodb-client.w" as dynamodb_client;
+
+interface Process {
+  inflight kill(): void;
+}
+
+struct SpawnOptions {
```
eladb, line 11: @Chriscbr any idea when this will be supported in the sdk? This is the 4th implementation I've seen of this code (and all of them have issues I am sure)

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+  new() {
+    let containerName = "winglibs-dynamodb-{util.uuidv4()}";
+
+    let state = new sim.State();
+    this.endpoint = state.token("endpoint");
+
+    new cloud.Service(inflight () => {
+      let port = Util.getPort();
+
+      let docker = Util.spawn(
```
eladb, line 48: Would it make sense to use containers.Workload here?

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+          }
+        },
+      );
+
+      return () => {
+        docker.kill();
+      };
+    });
+
+    // The host will be ready when the endpoint is set.
```
eladb, line 71: Why is this needed?

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+  tableName: str;
+
+  new(props: dynamodb_types.TableProps) {
+    this.host = Host.of(this);
+
+    let tableName = this.node.addr;
+    let state = new sim.State();
+    this.tableName = state.token("tableName");
+
+    new cloud.Service(inflight () => {
```
eladb, line 102: @Chriscbr please take a look at this- another use case for `sim.Resource` we can use to validate the APIs.

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  pub connection(): dynamodb_types.Connection {
+    return {
+      endpoint: this.host.endpoint,
+      tableName: this.tableName,
+    };
+  }
+
+  pub setStreamConsumer(handler: inflight (dynamodb_types.StreamRecord): void) {
```
eladb, line 171: Need some shutdown code here

eladb (changes requested):
Crazy project!

dynamodb/dynamodb.sim.w
```diff
@@ -0,0 +1,214 @@
...
+        });
+      }
+
+      util.waitUntil(() => {
+        try {
+          client.createTable({
+            TableName: tableName,
+            AttributeDefinitions: attributeDefinitions.copy(),
+            KeySchema: keySchemas.copy(),
+            BillingMode: "PAY_PER_REQUEST",
```
staycoolcall911, line 138: Should we add `pointInTimeRecovery`? https://github.com/winglang/wing/pull/5730

dynamodb/dynamodb-client.w
```diff
@@ -0,0 +1,196 @@
...
+    this.tableName = props.tableName;
+    this.client = Util.createDocumentClient({
+      region: props.region,
+      credentials: props.credentials,
+      endpoint: props.endpoint,
+    });
+  }
+
+  pub inflight delete(options: dynamodb_types.DeleteOptions): dynamodb_types.DeleteOutput {
+    let response = this.client.delete({
```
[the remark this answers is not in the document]
eladb: let's do it and see how it feels.

dynamodb/dynamodb-client.w
```diff
@@ -0,0 +1,196 @@
...
+  endpoint: str?;
+  region: str?;
+  credentials: Credentials?;
+}
+
+pub inflight class Client impl dynamodb_types.IClient {
+  inflight tableName: str;
+  inflight client: DocumentClient;
+
+  inflight new(props: ClientProps) {
```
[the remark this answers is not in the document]
skyrpex: The props must include the tableName, so aren't optional.

dynamodb/dynamodb-client.w
```diff
@@ -0,0 +1,196 @@
...
+  inflight scan(input: Json): Json;
+  inflight transactGet(input: Json): Json;
+  inflight transactWrite(input: Json): Json;
+  inflight update(input: Json): Json;
+}
+
+struct Credentials {
+  accessKeyId: str;
+  secretAccessKey: str;
+}
```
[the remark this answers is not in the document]
skyrpex: Yes, we need to pass fake credentials to the DynamoDB container in the sim, otherwise the client will try to gather the credentials from environment variables. For example, this is what I get:

```
The Wing Console is running at http://localhost:3000/
Error: The SSO session associated with this profile has expired. To refresh this SSO session run aws sso login with the corresponding profile.
   --> dynamodb-client.w:66:12
```

dynamodb/README.md
```diff
@@ -0,0 +1,61 @@
...
+    {
+      name: "id",
+      type: "S",
+    },
+  ],
+  hashKey: "id",
+);
+
+// Streams.
+table.setStreamConsumer(inflight (record) => {
```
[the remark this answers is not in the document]
skyrpex: Maybe. It'd make the implementation more complicated, though.

IMO we should stick to this raw implementation, and see if users actually need more advanced patterns.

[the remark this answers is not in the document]
skyrpex: DynamoDB always emits events for all records. Not sure if that's what you were asking.

dynamodb/dynamodb.sim.w
skyrpex: I guess, I just tried it and the resulting code is much cleaner, but I noticed that the containers doesn't get removed when I Ctrl+C the `wing it` process.

Maybe we can improve the containers winglib. The difference is that I spawn the docker container without detach and with the `--rm` option.

dynamodb/README.md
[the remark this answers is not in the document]
skyrpex: In that case, it would need to be implemented manually for the sim AFAIK. Maybe we could implement it at a later stage, with proper testing.

dynamodb/dynamodb.sim.w
skyrpex: I got BSODs about functions not getting resolved otherwise.

dynamodb/dynamodb.sim.w
skyrpex: I haven't encountered a situation where this keeps running. If the table container stops being available, it will end up halting the loop.

skyrpex:
@Chriscbr done! Had to add support for global secondary indexes :)

Didn't include https://github.com/winglang/wing/blob/main/examples/tests/sdk_tests/dynamodb-table/aws-dynamodb.test.w because I didn't see the point of it.

Chriscbr:
Sweet, thanks

dynamodb/dynamodb.sim.w
Chriscbr: Thanks for the reference point

dynamodb/dynamodb-client.w
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return unsafeCast(this.client.transactWrite({
+      TransactItems: transactItems.copy(),
+    }));
+  }
+
+  pub inflight scan(options: dynamodb_types.ScanOptions?): dynamodb_types.ScanOutput {
+    let input: MutJson = options ?? {};
+    input.set("TableName", this.tableName);
+    return unsafeCast(this.client.scan(input));
```
Chriscbr, line 108: Curious for my own understanding, how tricky is it to implement this method without `unsafeCast`?

Is `let input: MutJson = options ?? {}` a workaround for Winglang not allowing you set struct fields?

Chriscbr (approved):
I'm not intimately familiar with all the API surface area of DynamoDB so my review mostly focused on code quality. I added a few questions but nothing blocking.

dynamodb/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@cdktf/provider-aws": "^19.9.0",
+    "@winglibs/containers": "^0.0.22",
+    "get-port": "^7.0.0"
+  },
+  "devDependencies": {
+    "@cdktf/provider-helm": "^10.0.0",
+    "@cdktf/provider-kubernetes": "^11.2.0",
+    "@cdktf/provider-null": "^10.0.0",
+    "cdk8s-plus-27": "^2.7.84"
+  }
```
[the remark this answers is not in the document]
skyrpex: Yes, they are needed by the containers winglibs. Otherwise, it will just fail to run (on my machine, at least).

dynamodb/dynamodb-client.w
skyrpex: > It looks like the loop is broken by throwing an error

It is intended. There's a try catch later that ignore certain kind of errors, but throw otherwise.

> how tricky is it to implement this method without unsafeCast?

I guess we could add the proper return types to `DocumentClient` in `dynamodb/dynamodb-client.w` now that they return the original information, but the options still need the `TableName` to be added manually... Any ideas?

eladb (approved):

PR #104, opened 2024-03-06
chore(lock): fix README

ekeren:


Chriscbr:
@ekeren You will need to update package.json

PR #113, opened 2024-03-12
chore(vite): add violet color to vite custom resource

ainvoner:


skyrpex:
You'll have to edit the package.json's `version` manually, otherwise will fail to publish.

skyrpex (approved):

PR #117, opened 2024-03-12
feat: `bring cognito`

eladcon:


cognito/lib.test.w
```diff
@@ -0,0 +1,32 @@
...
+api.get("/hello", inflight (req) => {
+  log("Hello World {Json.stringify(req.headers)}");
+
+  return  {
+    status: 200
+  };
+});
+
+let auth = new l.Cognito(api);
+auth.get("/hello");
```
ekeren, line 18: Let's also add a post and put example here

ekeren (approved):

PR #122, opened 2024-03-13
feat(eventbridge)!: new eventbridge api

eladcon:
* Optional name property
* changed `subscribeFunction` to `onEvent`
* `putEvents` is now variadic


.github/workflows/eventbridge-release.yaml
```diff
@@ -29,9 +29,25 @@ jobs:
       - name: Pack
         run: wing pack
         working-directory: eventbridge
+      - name: Get package version
```
eladb, line 32: This workflow is generated so you can't change it manually

eladb (approved):

eladb:
Edited title

PR #123, opened 2024-03-13
feat: `bring jwt`

eladcon:


jwt/utils.mts
```diff
@@ -0,0 +1,35 @@
+import jwt, { SignOptions, VerifyOptions } from "jsonwebtoken";
+import jwksClient from "jwks-rsa";
+
+export const _sign = async (data: Record<string, any>, secret: string, options?: SignOptions) => {
+  return jwt.sign(data, secret, options);
+};
+
+export const _verifyWithSecret = async (token: string, secret: string, options?: VerifyOptions) => {
+  return jwt.verify(token, secret, options);
+};
```
eladb, line 10:
```suggestion
export const _verifyWithSecret = jwt.verify;
```

I think this should also work (and safe some typing).

You can also do something like this:

```js
exports.jwt = async () => jwt;
```

And then, in Wing:

```js
interface IJwt {
  sign(data: Record<string, any>, secret: str, options: SignOptions?): void;
  verify(token: str, getKey, ...)
}

class Utils {
  extern "./utils.mts"
  pub static inflight jwt(): IJwt;
}

inflight () => {
  Utils.jwt().sign(...);
};
```

(basically, if the API of the JavaScript library you are wrapping can be expressed using Wing's interfaces then you can just access it directly and there's no need to wrap it in JavaScript functions...)

eladb (approved):

PR #126, opened 2024-03-14
feat: `bring openai`

ShaiBer:
This is a very early version that contains just the most basic functionality of the openai JS library.

The tests are commented out because one cannot pass due to a bug and the other because I don't know how to provide it with credentials in the CI


openai/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "type": "git",
+    "url": "https://github.com/winglang/winglibs.git",
+    "directory": "openai"
+  },
+  "author": {
+    "email": "shaib@wing.cloud",
+    "name": "Shai Ber"
+  },
+  "license": "MIT",
+  "dependencies": {
```
staycoolcall911, line 15: Should be changed to `peerDependencies` see #5470
```suggestion
  "peerDependencies": {
```

ShaiBer: OK, so I'm also also updating the template

ShaiBer: Actually, it is not in the template, but are you sure that it needs to be peerdependecy? I don't see any other winglib with it

MarkMcCulloh: It should be, it's just not enforced right now https://github.com/winglang/wing/issues/5470

eladb (approved):

PR #127, opened 2024-03-14
fix(containers): ensure containers are destroyed when wing console is interrupted

skyrpex:
Use `docker --rm --detach` along with a random container name.


skyrpex:
Superseded by https://github.com/winglang/winglibs/pull/128.

PR #129, opened 2024-03-14
feat: `bring ssm`

eladcon:


ssm/README.md
```diff
@@ -0,0 +1,30 @@
...
+npm i @winglibs/ssm
+```
+
+## Usage
+
+```js
+bring expect;
+bring ssm;
+
+let param = new ssm.Parameter(name: "/test/param1", value: "value1");
```
eladb, line 21: Kind of feels like we should support the use case of just referencing an existing parameter.

I am not sure I fully understand the use case for defining the parameter value in the same Wing app that references it (what's the benefit of going through SSM here?).

```js
let bla = new ssm.ParameterReference(name: "bla");

inflight () => {
  log(bla.value());
};
```

eladb: But it seems like the value is a required prop of `ssm.Parameter`...

eladb: Anyway - I think we should support both use cases:

```js
bring ssm;

// with value
new ssm.Parameter(name: "my-param", value: "my-value");

// without value
new ssm.ParameterReference(name: "my-param");
```

Maybe both can implement `IParameter`:

```js
interface IParameter {
  inflight value(): str;
}
```

Or something like this.

PR #134, opened 2024-03-19
chore(openai): a few refactors

eladb:
Reduce the amount of JavaScript code to a minimum, which also reduces the difference between the mock version and the client. This increases test coverage.


ShaiBer:
Thanks! I added a couple of fixes to make the build pass, approving now

ShaiBer (approved):

PR #138, opened 2024-03-21
fix: descriptive name for test from checks lib

subh-cs:
fixes https://github.com/winglang/wing/issues/5109

for this code
```js
bring checks;

new checks.Check(inflight () => {}) as "one";
new checks.Check(inflight () => {}) as "two";

test "test name example" {}
```

now it look like this

image


checks/check.w
```diff
@@ -78,7 +78,7 @@ pub class Check impl ICheck {
     }
 
     if props?.testing ?? true {
-      new std.Test(handler) as "test";
+      new std.Test(handler) as "{this.node.id}Test";
```
eladb, line 81:
```suggestion
      new std.Test(handler) as "check {this.node.id}";
```

checks/check.w
```diff
@@ -78,7 +78,7 @@ pub class Check impl ICheck {
     }
 
     if props?.testing ?? true {
```
eladb, line 80: can you please add a test?

eladb (approved):

PR #140, opened 2024-03-21
fix(dynamodb): pack .mjs files

skyrpex:
Fixes https://github.com/winglang/wing/issues/6005

### DIFF
--- a/dynamodb/package.json
+++ b/dynamodb/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@winglibs/dynamodb",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "DynamoDB library for Wing",
   "author": {
     "name": "Cristian Pallarés",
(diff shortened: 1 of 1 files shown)


staycoolcall911 (approved):

dynamodb/package.json
```diff
@@ -11,6 +11,10 @@
     "url": "https://github.com/winglang/winglibs.git",
     "directory": "dynamodb"
   },
+  "files": [
+    "*.w",
+    "*.mjs"
+  ],
```
MarkMcCulloh, line 17: Wing should automatically be adding these as part of the CLI. *.w is already included, and *.mjs should be fixed upstream

skyrpex: We need to be careful with the defaults, though. We might end up including config files and other unrelated stuff (ie, eslint config)

skyrpex: Maybe we should discuss wing pack altogether? How does it differ from npm pack?

MarkMcCulloh: For now it's primarily an opinioned set of defaults/checks. Ideally we don't want to rely on the user knowing all about the various package.json setting they need to handle to publish something.

PR #141, opened 2024-03-21
fix(containers)!: migrate to `sim.Container`

eladb:
Use the shiny new `sim.Container` to implement the simulator version of `Workload`.

BREAKING CHANGE: `sources` is a glob string instead of an array of strings.

### DIFF
--- a/containers/helm.extern.d.ts
+++ b/containers/helm.extern.d.ts
@@ -1,0 +1,282 @@
+export default interface extern {
+  toHelmChart: (wingdir: string, chart: Chart) => string,
+}
+/** Trait marker for classes that can be depended upon.
+The presence of this interface indicates that an object has
+an `IDependable` implementation.
+
+This interface can be used to take an (ordering) dependency on a set of
+constructs. An ordering dependency implies that the resources represented by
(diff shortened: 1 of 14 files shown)


containers/helm.extern.d.ts
```diff
@@ -0,0 +1,282 @@
...
+  /** Returns all the included API objects. */
+  readonly apiObjects: (readonly (ApiObject)[]);
+  /** Generates a app-unique name for an object given it's construct node path.
+  Different resource types may have different constraints on names
+  (`metadata.name`). The previous version of the name generator was
+  compatible with DNS_SUBDOMAIN but not with DNS_LABEL.
+  
+  For example, `Deployment` names must comply with DNS_SUBDOMAIN while
+  `Service` names must comply with DNS_LABEL.
+
```
eladb, line 265: @MarkMcCulloh this definitely feels off.

MarkMcCulloh: What do you mean?

eladb: This generated .d.ts file.

(related to our discussion earlier about source of truth)

skyrpex (approved):

PR #144, opened 2024-03-24
feat(dynamodb): adding update and onInsert, onUpdate and onDelete method

marciocadev:
Addition of the update method and the filters onInsert, onUpdate, and onDelete to the event stream.

I also added tests for the new features.

It works for both `tf-aws` and `sim`


tsuf239:
those are failing on tf-aws with no explanation:
image

marciocadev:
> those are failing on tf-aws with no explanation: image

Tsuf, I found something strange when I started analyzing the problem; this was working before.

When I ran tests on AWS, I noticed that it was generating more Lambdas than it should, so I put together the code below to test:

```wing
import cloud;
import util;
import "../dynamodb.w" as dynamodb;

let table = new dynamodb.Table(
  attributes: [
    {
      name: "id",
      type: "S",
    },
  ],
  hashKey: "id",
);

new cloud.Function(inflight () => {
  table.put(
    Item: {
      id: "1",
      body: "hello",
    },
    ReturnValues: "NONE",
  );
});
```

In theory, it should only generate one Lambda on AWS, but it's generating 2 Lambdas: one with the function code and another called handler that I couldn't identify what it does.

image
image

I'm trying to understand why this is happening.

PR #147, opened 2024-03-25
fix(checks): latest does not check for nil

hasanaburayyan:
Maybe is a smell that we need `Struct.tryFromJson` but the core issue is the undefined object passed validation, which is surprising to me, but non the less there is some cleanup needed in Struct.fromJson since I made the changes to only return back the fields that match, and not the whole object.

### DIFF
--- a/checks/results.w
+++ b/checks/results.w
@@ -56,7 +56,10 @@ pub class Results {
   pub inflight latest(checkid: str): CheckResult? {
     let key = this.makeLatestKey(checkid);
     let s = this.bucket.tryGetJson(key);
-    return CheckResult.fromJson(s);
+    if let s = s {
+      return CheckResult.fromJson(s);
+    }
+    return nil;
   }
 
   inflight makeKey(checkid: str, key: str): str {
--- a/checks/package.json
+++ b/checks/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@winglibs/checks",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "description": "Cloud checks",
   "publishConfig": {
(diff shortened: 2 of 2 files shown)


checks/results.w
```diff
@@ -56,7 +56,10 @@ pub class Results {
   pub inflight latest(checkid: str): CheckResult? {
     let key = this.makeLatestKey(checkid);
     let s = this.bucket.tryGetJson(key);
-    return CheckResult.fromJson(s);
+    if let s = s {
+      return CheckResult.fromJson(s);
+    }
```
Chriscbr, line 61: Yeah - this feels like is a type checking bug in the compiler to me. If `s` is `Json?`, and `fromJson` only accepts `Json`, then the optional value shouldn't be accepted, I think.

(Struct.tryFromJson also sounds like a nice idea 👍 )

Chriscbr (approved):

checks/results.w
hasanaburayyan: Yea so first Im an idiot we already have `Struct.tryFromJson` 😁 and you are right this is strange, since `fromJson` expects `Json` so probably as you said some bug in type checking.

hasanaburayyan: Im able to reproduce with just:
```js
let f = (): Json? => {
  return nil;
};


struct Foo {
  name: str;
}

Foo.fromJson(f());
```

Going to open an issue

hasanaburayyan: Actually simpler example that compiles
```js
let f = (): Json? => {
  return nil;
};

let y = (x: Json) => {};

y(f());
```

PR #157, opened 2024-03-26
chore(dynamodb): change extern from .mjs to .ts

skyrpex:


dynamodb/dynamodb.ts
```diff
@@ -100,9 +100,9 @@ const processRecords = async (endpoint, tableName, handler, options) => {
     try {
       const { Streams } = await client.listStreams({ TableName: tableName });
       await Promise.all(
-        Streams.map(({ StreamArn }) =>
```
eladb, line 103: we should:

```ts
import type extern from "./dynamodb.extern";
```

And use the `: extern["foo"]` type annotations to enforce typing.

eladb (approved):

dynamodb/dynamodb.ts
skyrpex: Had to add some ts-ignore's to the DynamoDB clients.

eladb (approved):

PR #158, opened 2024-03-26
feat(dynamodb)!: api tweaks

eladb:
## `table.connection`

`table.connection` now includes `clientConfig` instead of just `endpoint`, so it can be used to create an AWS SDK client without any hacks.

I also changed it from a method to a property, so it can be lifted across inflight.

## `table.tableName`

Added a `tableName` property with the name of the table.

Found a bug in `tf-aws` where `name` was not respected.

Tested on AWS.

BREAKING CHANGE: `table.connection()` (method) is now `table.connection` (property). The `connection.endpoint` attribute is now `connection.clientConfig.endpoint`.

### DIFF
--- a/dynamodb/README.md
+++ b/dynamodb/README.md
@@ -51,9 +51,16 @@ test "put and query" {
 In case you want to instantiate your own DynamoDB SDK, you can get the connection details like this:
 
 ```wing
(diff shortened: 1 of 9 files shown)


skyrpex (approved):
Great!

PR #174, opened 2024-04-01
feat: `bring functools`

Chriscbr:
WIP. The tests aren't passing yet, some kind of issue with the simulator and the dynamodb winglib, but just sharing this draft for visibility


Chriscbr:
Alternate name ideas - `bring fntools`, `bring funcutil`, `bring fnutil`, `bring idempotency`...

functools/README.md
```diff
@@ -0,0 +1,63 @@
...
+Lambda](https://docs.powertools.aws.dev/lambda/typescript/latest/utilities/idempotency/),
+provides a simple solution to convert your inflight functions into idempotent
+operations which are safe to retry. Under the hood, it extends `cloud.Function` to communicate with a persistence layer to prevent handlers from executing
+more than once on the same event payload, and it ensures handlers returns the
+same result when called with the same payload.
+
+```js
+bring functools;
+
+let submitOrder = new functools.IdempotentFunction(inflight (order: Json) => {
```
eladb, line 27: It is possible to design this API so it can be composed with other `cloud.Function` "extensions" instead of subclass?

Can we get to something like this?

```js
let f = new cloud.Function(...);

funtools.addIdempotency(f);
```

Chriscbr: > Can we get to something like this?

It's possible, though it would require some form of inversion of control or middleware mechanism, eg

```js
let f = new cloud.Function(...);
f.useMiddleware(inflight (handler) => {
  return inflight (payload) => {
    // insert wrapper logic...
    let result = handler(payload);
    return result;
  };
});
funtools.addIdempotency(f);
```

But it's a bit restrictive because the inflight handler would still need to be `inflight (str?): str?` at the end of the day

eladb: So how would you recommend we compose multiple functools?

Chriscbr: I guess it depends, what kinds of other functools do you have in mind ? :) addMetrics? addTelemetry? addCircuitBreaker?

The middleware mechanism could probably work alright - if you want to take a stab at implementing something, I'd say go for it 👍  it seems pretty powerful.

Also `funtools` could be a nice name...

marciocadev:
Be careful with the use of cloud.Table for idempotency, @Chriscbr.
In AWS, cloud.Table is implemented using DynamoDB, which allows for atomic transactions.
However, on other platforms, if implemented with a feature that does not allow this type of transaction, idempotency may be compromised.

PR #178, opened 2024-04-03
fix(dynamodb): ensure table is ready

skyrpex:
Use the DescribeTable operation on a table to ensure that the table is ready.

### DIFF
--- a/dynamodb/dynamodb.sim.w
+++ b/dynamodb/dynamodb.sim.w
@@ -171,10 +171,14 @@ pub class Table_sim impl dynamodb_types.ITable {
           // Wait until we can describe the table. This is
           // to ensure that the table is ready to be used.
           util.waitUntil(() => {
-            client.describeTable({
-              TableName: tableName,
-            });
-            return true;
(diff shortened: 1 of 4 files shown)


dynamodb/dynamodb.sim.w
```diff
@@ -167,6 +168,15 @@ pub class Table_sim impl dynamodb_types.ITable {
             });
           }
 
+          // Wait until we can describe the table. This is
+          // to ensure that the table is ready to be used.
+          util.waitUntil(() => {
+            client.describeTable({
```
eladb, line 174: If the table is not ready, this will throw. I believe you need to try/catch and return false.

eladb (changes requested):

Chriscbr (approved):

dynamodb/dynamodb.sim.w
Chriscbr, line 1: You'l have to bump package.json until we're able to automate it

eladb (approved):

PR #179, opened 2024-04-03
chore: use peer dependencies and make packages more similar

MarkMcCulloh:
The primary goal of this PR is to change all package to use peerDependencies instead of dependencies (see https://github.com/winglang/wing/issues/5470)

- It was necessary to update cdktf and some cdktf providers to avoid peers issues. This lead to various other updates via lockfile updates.
- To further simplify packages, removed any usages of "winglang" as a dependency.
- Commited extern .d.ts files
- Removed some extraneous dependencies (peer deps everywhere _should_ help with needing to include transitive deps)


postgres/package.json
```diff
@@ -12,14 +12,15 @@
...
-    "pg": "^8.11.3"
+    "@cdktf/provider-aws": "^19.12.0",
+    "@winglibs/containers": "^0.0.26",
+    "pg": "^8.11.3",
+    "cdktf": "^0.20.6"
+  },
+  "overrides": {
+    "@cdktf/provider-aws": "$@cdktf/provider-aws",
+    "cdktf": "$cdktf"
   },
```
MarkMcCulloh, line 24: Temporary fix! `@winglibs/containers` dependency has peer conflicts since it still uses older versions of these deps. A PR will fast follow this one to update any packages that use `@winglibs/containers`

containers/workload.sim.w
```diff
@@ -37,6 +37,7 @@ pub class Workload_sim {
       });
 
       let s1 = new cloud.Service(inflight () => {
+        log("waiting for container to be ready...");
```
Chriscbr, line 40: Debugging?

Chriscbr (approved):

budget/package.json
```diff
@@ -17,16 +17,13 @@
   },
   "license": "MIT",
   "peerDependencies": {
-    "cdktf": "*",
-    "winglang": "*"
+    "cdktf": "^0.20.6",
```
eladb, line 20: can we add some validation in the PR builds in this repo?

PR #182, opened 2024-04-04
feat: `bring react`

meirdev:


reactapp/utils.js
```diff
@@ -0,0 +1,60 @@
...
+
+  // proc.stdout.setEncoding("utf8");
+  // proc.stdout.on("data", (data) => {
+  //   console.log(data);
+  // });
+
+  // proc.stderr.setEncoding("utf8");
+  // proc.stderr.on("data", (data) => {
+  //   console.log(data);
+  // });
```
Chriscbr, line 29: The information from the `dev` or `build` command could be useful for debugging. I was trying out the library and got stuck momentarily because I didn't realize I forgot to run `npm install` in the react app folder. 😅 

Maybe it's worth at least logging the stderr by default:
```suggestion
  // proc.stdout.setEncoding("utf8");
  // proc.stdout.on("data", (data) => {
  //   console.log(data);
  // });

  proc.stderr.setEncoding("utf8");
  proc.stderr.on("data", (data) => {
    console.log(`[${command}] ${data}`);
  });
```

reactapp/test/main/main.w
```diff
@@ -0,0 +1,20 @@
...
+  };
+});
+
+let project = new reactapp.ReactApp(
+  projectPath: "../react-project",
+  localPort: 4500,
+);
+
+project.addEnvironment("API_URL", api.url);
+project.addEnvironment("TEXT", "Hello World!");
```
Chriscbr, line 20: Maybe we can add a test, like:
```suggestion
project.addEnvironment("TEXT", "Hello World!");

if util.env("WING_TARGET") == "sim" {
  test "we can access the app" {
    let url = project.url;
  
    util.waitUntil(inflight () => {
      try {
        let response = http.get(url);
        return response.status == 200;
      } catch e {
        return false;
      }
    }, timeout: 10s);
  
    let response = http.get(url);
    expect.equal(response.status, 200);
    assert(response.body.startsWith("<!DOCTYPE html>"));
  }
}


```

Chriscbr (approved):
Nice work 🙌

reactapp/test/main/main.w
meirdev: I need to run `npm install` in the `react-project` dir before the test starts running, but when I add a new step in `reactapp-pull.yaml`:

```yaml
      - name: Install React project dependencies
        run: npm install
        working-directory: reactapp/test/react-project
```

the Pull Request Diff workflow failed :(

reactapp/README.md
```diff
@@ -0,0 +1,95 @@
+# ReactApp
```
eladb, line 1: Let's call this library `react` and then:

```
bring react;

new react.App(...);
```

react/README.md
```diff
@@ -0,0 +1,95 @@
...
+    <div className="App">
+      <p>TEXT: {TEXT}</p>
+      <p>FROM API: {text}</p>
+    </div>
+  );
+}
+
+export default App;
+```
+
```
eladb, line 92: Please add a "maintainers" section

react/lib.w
```diff
@@ -0,0 +1,41 @@
...
+      throw "unknown platform {target}";
+    }
+
+    this.url = this.platform.getUrl();
+
+    new ui.Field("url", inflight () => {
+      return this.url;
+    }, link: true);
+
+    if target == "sim" {
```
eladb, line 31: Why only for `sim`? The endpoint is also relevant in production

react/lib.w
```diff
@@ -0,0 +1,41 @@
...
+bring util;
+
+bring "./shared.w" as appShared;
+bring "./sim.w" as appSim;
+bring "./tf-aws.w" as appTfAws;
+
+pub class App {
+  pub url: str;
+
+  platform: appShared.IApp;
```
eladb, line 12:
```suggestion
  inner: appShared.IApp;
```

We usually call this `inner`

react/shared.w
eladb, line 1: We normally call this file `api.w`

react/shared.w
```diff
@@ -0,0 +1,51 @@
...
+  protected path: str;
+
+  protected env: MutMap<str>;
+
+  protected startCommand: str;
+  protected buildCommand: str;
+  protected buildDir: str;
+
+  new(props: AppPros) {
+    this.path = fs.absolute(nodeof(this).app.entrypointDir, props.projectPath);
```
eladb, line 32: @Chriscbr this is going to bite us in the A because we don't have something like `__dirname` yet.

So for example, if you create a library or a module in a subdirectory and want to put your project in an adjacent directory, this will not work.

react/shared.w
```diff
@@ -0,0 +1,51 @@
+bring cloud;
+bring fs;
+
+bring "./utils.w" as utils;
+
+pub struct AppPros {
```
eladb, line 6: Please add some docstrings here

react/sim.w
```diff
@@ -0,0 +1,51 @@
...
+
+      fs.writeFile(
+        fs.join(this.path, "public", "wing.js"),
+        "// This file is generated by wing\nwindow.wingEnv = {Json.stringify(this.env)};",
+      );
+
+      this.env.set("PORT", "{port}");
+
+      if this.props.useBuildCommand? == true {
+        utils.Utils.exec(this.buildCommand, this.env, this.path);
```
eladb, line 34: I believe we have `util.shell`

eladb (changes requested):

react/lib.w
meirdev: `cloud.Website` already has `cloud.Endpoint`

eladb (approved):

PR #189, opened 2024-04-09
chore: add test retries

Chriscbr:


.github/workflows/canary.yaml
```diff
@@ -29,9 +29,14 @@ jobs:
...
-      - name: Test
-        run: wing test
-        working-directory: bedrock
+      - name: Run tests
+        uses: nick-fields/retry@v3
+        with:
+          max_attempts: 3
+          command: |-
+            cd bedrock
+            wing test
```
eladb, line 38: I believe we have built-in retry support in `wing test`, no?

Chriscbr: Yep. But each library is allowed to have a custom test script (for example, the tests for `bring react` install dependencies in the react website folder first) so using this retry action will retry everything instead of just `wing test`

eladb: How exactly? It seems like it's just running `wing test` inside...

eladb: Oh I see! 

Okay

PR #191, opened 2024-04-09
feat: `bring python`

eladcon:


python/README.md
```diff
@@ -0,0 +1,62 @@
+# python
+
+A Wing library for integrating with Python code.
```
eladb, line 3:
```suggestion
A Wing library for running Python code in inflight.
```

python/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```js
+// main.w
+let bucket = new cloud.Bucket();
+let func = new python.Function(
+  path: "./test-assets",
+  handler: "main.handler",
+);
+
+func.liftClient("bucket", bucket, ["get", "put"]);
```
eladb, line 40: I think `lift()` should be sufficient:

```suggestion
func.lift("bucket", bucket, ["get", "put"]);
```

I think we can also use keywords to make this nicer:

```js
func.lift(bucket, as: "my_bucket", allow: ["get", "put"]);
```

What do you think?

python/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+```python
+# main.py
+from wing_helper import *
```
eladb, line 45:
```suggestion
from wing import *
```

The `_helper` is not adding any context (and actually "from wing import *" reads so nice).

python/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+def handler(event, context):
+  client = get_client("bucket")
```
eladb, line 48:
```suggestion
  client = lift("bucket")
```

I would also call this `lift()` or `lifted()`

python/README.md
```diff
@@ -0,0 +1,62 @@
...
+  client.put("test.txt", "Hello, world!")
+  
+  return {
+    "statusCode": 200,
+    "body": "Hello!"
+  }
+```
+
+#### Supported Wing Resource:
```
eladb, line 57: :-)

python/README.md
```diff
@@ -0,0 +1,62 @@
...
+  let res = func.invoke();
+}
+```
+
+It is also possible to interact with Wing resources through the python code
+
+```js
+// main.w
+let bucket = new cloud.Bucket();
+let func = new python.Function(
```
eladb, line 35: How would this work with something like e.g. `cloud.Api`?

Ideally this would have been a `python.Inflight` instead of `python.Function` and then maybe I could do:

```js
let api = new cloud.Api();
api.post("/users", python.Inflight(...));
```

Not sure how this would work but that would be my ideal DX, no?

Or event `python.inflight()` (all lowercase)

python/sim/containers.w
```diff
@@ -0,0 +1,259 @@
+bring http;
+bring util;
+bring cloud;
+bring sim;
+bring fs;
+
+pub class Util {
+  extern "./util.js" pub static inflight shell(command: str, args: Array<str>, pathEnv: str, cwd: str?): str;
```
eladb, line 8: don't we have `util.shell()`?

python/sim/containers.w
```diff
@@ -0,0 +1,259 @@
...
+  extern "./util.js" pub static contentHash(files: Array<str>, cwd: str): str;
```
eladb, line 9: I think we have `fs.md5()`.

python/sim/containers.w
```diff
@@ -0,0 +1,259 @@
...
+  extern "./util.js" pub static dirname(): str;
```
eladb, line 10: ahhhhhhhh @Chriscbr is this prioritized?

python/sim/containers.w
```diff
@@ -0,0 +1,259 @@
...
+   * source files.
+   */
+  sourceHash: str?;
+}
+
+pub struct WorkloadProps extends ContainerOpts {
+
+}
+
+pub class Workload_sim {
```
eladb, line 66: what's this about?

python/sim/function.w
```diff
@@ -0,0 +1,92 @@
...
+
+    let outdir = libutil.build(
+      entrypointDir: entrypointDir,
+      workDir: workDir,
+      path: props.path,
+      homeEnv: homeEnv,
+      pathEnv: pathEnv,
+    );
+
+    let runner = new containers.Workload_sim(
```
eladb, line 35: why not use `sim.Container`?

python/sim/util.js
```diff
@@ -0,0 +1,48 @@
...
+        console.error(stderr);
+        return reject(error);
+      }
+
+      return resolve(stdout ? stdout : stderr);
+    });
+  });
+};
+
+exports.entrypointDir = function (scope) {
```
eladb, line 24: `nodeof(this).app.entrypointDir` should do it

eladb (changes requested):

python/README.md
ekeren: I like this API suggestion. 
@eladcon, once you move forward lets also make these API  changes to tsoa

python/README.md
eladb: Yes I understand the problem...

Generics...

Can we do something like this?

```js
python.Api.inflight()
python.Function.inflight()
```

python/test-assets/wing.py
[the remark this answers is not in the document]
ekeren: yep makes sense

python/lib.w
ekeren, line 1: Do you plan to support for consuming queue events as well? 

```ts
let queue = new cloud.Queue();
queue.setConsumer(python.QueueConsumer.Inflight(this,
  path: "./test-assets",
  handler: "main.handler"
).lift(bucket, as: "bucket", allow: ["get", "put"]));
```

python/lib.test.w
```diff
@@ -0,0 +1,17 @@
+bring cloud;
+bring http;
+bring expect;
+bring "./lib.w" as python;
+
+let bucket = new cloud.Bucket();
+let func = new cloud.Function(new python.InflightFunction(
+  path: "./test-assets",
+  handler: "main.handler"
+).lift("bucket", bucket, allow: ["get", "put"]));
```
eladb, line 10:
```suggestion
let func = new cloud.Function(new python.InflightFunction(
  path: "./test-assets",
  handler: "main.handler",
  lift: {
    bucket: { obj: bucket, allow: ["get", "put"] }
  }
));
```

Maybe we can also support specifying lifts in props (in addition to the `lift()` method).

eladb (approved):

PR #192, opened 2024-04-09
feat(postgres)!: public connection string, change default postgres version to 16

ekeren:
Done @skyrpex and @ekeren 

- Replaced `container.Workload` with `sim.container`
- Exposed client configuration 
- Upgrade default postgres version to 16

### DIFF
--- a/postgres/lib.w
+++ b/postgres/lib.w
@@ -1,12 +1,21 @@
 bring aws;
 bring cloud;
-bring containers;
 bring http;
 bring util;
+bring sim;
 bring "constructs" as constructs;
 bring "cdktf" as cdktf;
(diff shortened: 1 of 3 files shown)


postgres/lib.w
```diff
@@ -1,13 +1,22 @@
...
-bring containers;
 bring http;
 bring util;
+bring sim;
 bring "constructs" as constructs;
 bring "cdktf" as cdktf;
 bring "@rybickic/cdktf-provider-neon" as rawNeon;
 bring "@cdktf/provider-aws" as tfaws;
 
+pub struct ConnectionOptions {
```
eladb, line 11: I'd just call this `Connection` or `Address`

eladb (commented):
Test?

PR #205, opened 2024-04-21
fix(tsoa): server port keeps changing between compilations

eladcon:
saving the last used port in a bucket so it could be used for later compilations

### DIFF
--- a/tsoa/app.js
+++ b/tsoa/app.js
@@ -1,7 +1,8 @@
+const { Socket } = require("net");
 const express = require("express");
 const { setLifted } = require("./clients");
 
-exports.runServer = async (routes, clients) => {
+exports.runServer = async (routes, clients, lastPort) => {
   const app = express();
 
   app.use((req, res, next) => {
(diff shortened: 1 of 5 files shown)


ekeren:
Making sure you have tests as a use case on our plate

tsoa/sim.w
```diff
@@ -27,6 +28,7 @@ pub class Service_sim impl types.IService {
     }
 
     this.state = new sim.State();
+    this.bucket = new cloud.Bucket();
```
staycoolcall911, line 31: Just curious (not sure it matters):
Why put it into a new bucket? Why not use the `sim.State`?

tsoa/app.js
```diff
@@ -1,7 +1,8 @@
+const { Socket } = require("net");
 const express = require("express");
 const { setLifted } = require("./clients");
 
-exports.runServer = async (routes, clients) => {
+exports.runServer = async (routes, clients, lastPort) => {
```
staycoolcall911, line 5: Can we add a test for that?

staycoolcall911 (approved):
LGTM!

PR #207, opened 2024-04-22
feat(postgres): retain postgres database state

Chriscbr:
Fixes https://github.com/winglang/wing/issues/6284

Depends on changes from https://github.com/winglang/wing/pull/6295, https://github.com/winglang/wing/pull/6299, and https://github.com/winglang/wing/pull/6336


postgres/lib.w
```diff
@@ -66,7 +66,8 @@ class DatabaseSim impl IDatabase {
       env: {
         POSTGRES_PASSWORD: "password"
       },
-      containerPort:5432
+      containerPort: 5432,
+      volumes: ["$WING_STATE_DIR:/var/lib/postgresql/data"],
```
ekeren, line 70: Maybe add a platform parameters to control the default

Chriscbr: Would it help if the Wing Console had a way to reset an app's state?

Chriscbr: Feel free to open a ticket and I can add a platform parameter for the postgres lib if you think there are use cases

PR #211, opened 2024-04-24
fix: lifted mutables become immutable

yoav-steinberg:
Fixes broken test and libs due to https://github.com/winglang/wing/pull/6303
* Lifted mutables become immutable, so we can't treat them an mutable.
* Don't used `std.Node` it's not liftable - need to consider what to do with it, the correct way is to use the builtin `nodeof()`.


Chriscbr (approved):

eladb:
This is very cool

PR #223, opened 2024-05-06
feat(containers): helm output

eladb:
Introduce an experimental parameter `provider` in `wing.toml` to allow specifying the backend for containers.

Also - fix how readiness checks are implemented.


containers/wing.toml
```diff
@@ -0,0 +1,2 @@
+[containers]
+provider="eks"
```
MarkMcCulloh, line 2: Isn't this basically what a platform is? Having this as a parameter is kinda confusing in our current model

containers/tfaws-eks.w
```diff
@@ -23,8 +22,6 @@ interface ICluster extends std.IResource {
 pub class ClusterBase impl ICluster {
   pub attributes(): ClusterAttributes { 
     throw "Not implemented"; 
-    // WORKAROUND: Compiler doesn't recognize that this will never return, so we need to return something
-    return { certificate: "", endpoint: "", name: "" };
```
Chriscbr, line 27: 😄

containers/tfaws-eks.w
```diff
@@ -81,37 +78,36 @@ class ClusterRef extends ClusterBase impl ICluster {
...
     return existing ?? newCluster();
   }
 
-
-  static tryGetClusterAttributes(): ClusterAttributes? {
-    if !values.Values.has("eks.cluster_name") {
-      return nil;
+  static tryReadParameters(scope: std.IResource): ClusterAttributes? {
+    if let json = nodeof(scope).app.parameters.value("eks") {
+      log(json);
```
Chriscbr, line 104: debugging?

containers/helm.w
```diff
@@ -0,0 +1,87 @@
...
+
+    let deployment = new plus.Deployment(
+      replicas: props.replicas,
+      metadata: {
+        name: props.name
+      },
+    );
+
+    deployment.addContainer(
+      image: "\{\{ .Values.image }}",
```
Chriscbr, line 40:
```suggestion
      image: r"{{ .Values.image }}",
```

Chriscbr (approved):

containers/wing.toml
eladb: I agree but as long as we don't have a way to inject dependencies for non-SDK classes (https://github.com/winglang/wing/issues/4930) we need something like this in order to be able to configure the behavior of `containers.Workload`.

containers/tfaws-eks.w
eladb: yes

PR #225, opened 2024-05-08
chore: improve tests-all.sh

yoav-steinberg:
* Can specify specific wing binary to use for tests
* Explicitly use `bash` so `pipefail` will work
* Use custom `test.sh` when avaialble (for cdk8s)


Chriscbr (approved):
🙏

PR #229, opened 2024-05-11
feat: bring slack

hasanaburayyan:


slack/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```sh
+npm i @winglibs/slack
+```
+## Bring it
+
+```js
+bring slack;
+
+let slackbot = new slack.Slackbot();
```
eladb, line 21:
```suggestion
let slackbot = new slack.SlackApp();
```

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+  ignoreBots: bool;
+  botToken: cloud.Secret;
+
+  new(props: SlackbotProps) {
+    this.onEventHandlers = MutMap<inflight (context.CtxEventCallBack, Json): void>{};
+    this.ignoreBots = props?.ignoreBots ?? true;
+    this.botToken = new cloud.Secret(name: "{this.node.id}_botToken");
+    this.api = new cloud.Api();
+
+    if util.env("WING_TARGET") == "tf-aws" {
```
eladb, line 29: Technically this is harmless in the simulator so the if statement is not really needed.

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+      new cdktf.TerraformOutput(value: "{this.api.url}/slack/events", description: "Slack Request URL");
+    }
+
+    this.api.post("/slack/events", inflight (req) => {
+      let eventRequest = events.SlackEvent.fromJson(Json.parse(req.body!));
+      if eventRequest.type == "url_verification" {
+        let verificationEvent = events.VerificationEvent.fromJson(Json.parse(req.body!));
```
eladb, line 36:
```suggestion
        let verificationEvent = events.VerificationEvent.parseJson(req.body!);
```

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+        return {
+          status: 200,
+          body: Json.stringify({
+            challenge: verificationEvent.challenge
+          })
+        };
+      }
+
+      if eventRequest.type == "event_callback" {
+        let callBackEvent = events.CallBackEvent.fromJson(Json.parse(req.body!)["event"]);
```
eladb, line 46: Use parseJson ?

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+
+        if this.ignoreBots {
+          if let bot_id = callBackEvent.bot_id {
```
eladb, line 49:
```suggestion
          if callBackEvent.bot_id == nil {
```

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+    });
+  }
+
+  /// Register an event handler
+  pub onEvent(eventName: str, handler: inflight(context.CtxEventCallBack, Json): void) {
+    this.onEventHandlers.set(eventName, handler);
+  }
+
+  /// Retrieve a channel object from a channel Id
+  pub inflight getChannel(channelId: str): context.Channel {
```
eladb, line 69: We usually omit the "get"

slack/message.w
```diff
@@ -0,0 +1,49 @@
+// TODO: right now there is no way to define how an enum
```
eladb, line 1: Do we have a tracking issue?

slack/message.w
```diff
@@ -0,0 +1,49 @@
...
+    this.sections = MutArray<Section>[];
+  }
+
+  pub addSection(section: Section) {
+    this.sections.push(section);
+  }
+
+  /// Returns string representation of message for 
+  /// over the wire transmission
+  pub buildMessage(): str {
```
eladb, line 39:
```suggestion
  pub render(): str {
```

eladb, line 39:
```suggestion
  pub render(): str {
```

Or would it make sense to return a Json? And then call this `toJson`?

slack/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@winglibs/slack",
+  "version": "0.1.0",
```
eladb, line 3: Missing description

slack/util.w
```diff
@@ -0,0 +1,15 @@
+bring http;
+
+pub class SlackUtils {
+  pub static inflight post(body: Json, botToken: str) {
```
eladb, line 4: Can't this be a private method of `slack.App`?

eladb (changes requested):
Missing tests...


I think this library needs some special behavior when running inside a test. Maybe some kind of mocking capabilities like we implemented for Bedrock

slack/README.md
```diff
@@ -0,0 +1,91 @@
...
+6. Copy the Bot User OAuth Token
+7. Navigate `Event Subscriptions` and enable events, then subscribe to bot events:
+  - app_mention
+
+### Configuring Application Secrets
+
+Before your bot can connect to Slack, you need to configure the required bot token as a secret in your project. You can do this by simply running the following command:
+
+```sh
+wing secrets -t PLATFORM_TARGET
```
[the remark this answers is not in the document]
hasanaburayyan: Or maybe it just makes more sense to repeat the step in both "running in sim" and "running in aws"

slack/context.w
```diff
@@ -0,0 +1,93 @@
+bring "./message.w" as msg;
+bring "./util.w" as helpers;
+bring "./events.w" as events;
+
+/// The bahvioral interface of a sender
+pub inflight interface ISender {
```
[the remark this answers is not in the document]
hasanaburayyan: Im game

slack/events.w
```diff
@@ -0,0 +1,23 @@
...
+  type: str;
+}
+
+pub struct CallBackEvent {
+  user: str;
+  type: str;
+  ts: str;
+  team: str;
+  channel: str;
+  event_ts: str;
```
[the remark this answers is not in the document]
hasanaburayyan: @eladb these have to be spelled this exact way, until we have support for https://github.com/winglang/wing/issues/3686

slack/lib.w
```diff
@@ -0,0 +1,72 @@
...
+bring util;
+bring ui;
+
+/// Properties for Slack bot
+pub struct SlackbotProps {
+  ignoreBots: bool?;
+}
+
+
+pub class Slackbot {
```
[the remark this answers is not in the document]
hasanaburayyan: Yea I guess in Slack a bot is a subset of a slack app. So it makes sense to call it an app

slack/context.w
```diff
@@ -0,0 +1,93 @@
...
+  inflight postText(message: str): void;
+}
+
+/// Represents the context of a slack channel
+pub inflight class Channel impl ISender {
+  /// The channel id
+  pub channel_id: str;
+  botToken: str;
+  
+  new (channel: str, botToken: str) {
```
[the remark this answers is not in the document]
hasanaburayyan: Originally that was my plan, but im not sure if Im dealing with a bug or limitation. Essentially within the postMessage() methods when trying to use `cloud.Secret` Ill receive an error that tells me I need to use `lift()` explicitly however, when I try to do an explicit lift I cant because the secret is an inflight object (since the class is an inflight class) 

Ill try and play with this a bit more to understand whats going on a bit better. Might call out for some help from our lifting experts.

slack/events.w
```diff
@@ -0,0 +1,23 @@
+pub struct SlackEvent {
+  type: str;
+}
+
+pub struct CallBackEvent {
```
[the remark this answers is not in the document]
hasanaburayyan: So in the outter scope of a slack event there exists two types of typically its an `event_callback` but there is also a `url_verification` 

Then there is an inner type of event which can be ALOT: https://api.slack.com/events which was the motivation for being specific with the naming

slack/util.w
hasanaburayyan: No since this is needed by context objects `Channel` and `Thread` I wanted to make like an abstract context class that `Channel` and `Thread` can inherit from to get this method, but alas we dont support that.

slack/context.w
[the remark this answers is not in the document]
hasanaburayyan: Had a chat with @Chriscbr looks like the core issue to passing the `cloud.Secret` around is related to: https://github.com/winglang/wing/issues/2730

I think for now its fine to proceed passing the token around as a string since its never exposed anywhere. Ill add a TODO: to revisit once we complete the mentioned issue (its part of stability milestone)

slack/message.w
hasanaburayyan: I did not find one, opened: https://github.com/winglang/wing/issues/6470

hasanaburayyan:
Ah shit, I forgot tests... adding...

slack/README.md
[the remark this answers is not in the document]
eladb: I would repeat for both steps because it is needed, separately, for each target.

slack/events.w
```diff
@@ -0,0 +1,23 @@
...
+  type: str;
+  ts: str;
+  team: str;
+  channel: str;
+  event_ts: str;
+  bot_id: str?;
+  app_id: str?;
+}
+
+pub struct MessageCallBackEvent extends CallBackEvent {
```
Chriscbr, line 16: Super nitpick (I'm guessing "callback" is meant in this sense? https://en.wikipedia.org/wiki/Callback_(computer_programming))
```suggestion
pub struct MessageCallbackEvent extends CallbackEvent {
```

slack/lib.w
```diff
@@ -0,0 +1,91 @@
...
+          if this.isTest {
+            return {
+              status: 200,
+              body: Json.stringify(handler(new context.EventContext_Mock(Json.parse(req.body!), ""), Json.parse(req.body!)))
+            };
+          } else {
+            // TODO: pass bot token as cloud.Secret rather than str once: https://github.com/winglang/winglibs/pull/229 is complete
+            return {
+              status: 200,
+              body: Json.stringify(handler(new context.EventContext(Json.parse(req.body!), this.botToken.value()), Json.parse(req.body!)))
```
Chriscbr, line 71: Can we shorten the code by calling `Json.parse(req.body!)` once earlier in the function and reusing the result everywhere?

slack/lib.w
```diff
@@ -0,0 +1,91 @@
...
+            };
+          }
+        }
+      }
+    });
+  }
+
+  /// Register an event handler
+  pub onEvent(eventName: str, handler: inflight(context.EventContext, Json): Json?) {
```
Chriscbr, line 80: What's the difference between the two arguments passed to the handler? It might be worth documenting somewhere (not sure)

Chriscbr (approved):

slack/events.w
hasanaburayyan: Yea you are right here.

slack/lib.w
hasanaburayyan: Basically there `EventContext` is the context in which the event happened, it gives access to clients for the channel and thread of event, whereas the second argument is the raw event itself.

I can add some doc details in README

slack/README.md
```diff
@@ -0,0 +1,121 @@
...
+```
+## Bring it
+
+```js
+bring slack;
+bring cloud;
+
+let myBotToken = new cloud.Secret(name: "mybot_token");
+
+let slackbot = new slack.App(botToken: myBotToken);
```
eladb, line 24: Sorry for the hassle. I am now realizing this is called a bot in the slack api, so we might as well call this `slack.Bot`

Sorry!

slack/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+First lets configure our Slack bot token as a secret in the simulator. 
+
+```sh
+wing secrets
+```
+
+When prompted, paste the Bot User OAuth Token you copied earlier.
+
+Next when running in the Wing Simulator, you will need to expose the endpoint of the Slackbot API, this can be done through the simulator console by selecting `Open tunnel for this endpoint`
```
eladb, line 67: Aren't we using websockets? Why do we need to expose?

slack/README.md
```diff
@@ -0,0 +1,121 @@
...
+If you want to post directly to a channel, you can do so by using the following code:
+
+```js
+bring slack;
+bring cloud;
+
+let slackbot = new slack.Slackbot();
+
+let postMessage = new cloud.Function(inflight () => {
+  let channel = slackbot.getChannel("C072T52EZ3Q");
```
eladb, line 117: Add some doc about where do I get the channel code?

Can I use the channel name?

slack/context.w
```diff
@@ -0,0 +1,145 @@
+bring "./message.w" as msg;
+bring "./util.w" as helpers;
+bring "./events.w" as events;
+
+/// The bahvioral interface of a thread
+pub inflight interface IThread {
+  inflight postMessage(message: msg.Message): Json;
```
eladb, line 7:
```suggestion
  postMessage(message: msg.Message): Json;
```

Since the entire interface is inflight I believe it shouldn't be needed to specify this for each method. No?

slack/context.w
```diff
@@ -0,0 +1,145 @@
...
+        channel: this.channel.id,
+        thread_ts: this.timestamp,
+        text: message
+      },
+      this.botToken
+    );
+  }
+}
+
+pub inflight class Channel_Mock extends Channel {
```
eladb, line 84: Why is this public?

Why the underscore in the name?

I haven't seen any documentation about this.

slack/context.w
```diff
@@ -0,0 +1,145 @@
...
+      body: "Totally sent that message to the thread :)"
+    };
+  }
+}
+
+/// Represents the context of an event callback
+pub inflight class EventContext {
+  pub thread: Thread;
+  pub channel: Channel;
+  new (rawContext: Json, botToken: str) {
```
eladb, line 128:
```suggestion
  new(rawContext: Json, botToken: str) {
```

eladb: Fix everywhere

slack/lib.test.w
```diff
@@ -0,0 +1,79 @@
...
+bring "./lib.w" as slack;
+bring "./message.w" as msg;
+bring "./events.w" as events;
+
+let token = new cloud.Secret();
+let app = new slack.App(botToken: token);
+
+app.onEvent("app_mention", inflight (ctx, event) => {
+  // Have the call just return the response from posting to thread (for testing)
+  let res = ctx.thread.post("message");
```
eladb, line 13: Maybe test that `event` is right

eladb (approved):

slack/README.md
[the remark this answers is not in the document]
hasanaburayyan: Technically an app can use a bot token or a user token and there are differences in what each can do iirc, but maybe this distinction does not make sense to introduce yet

slack/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+```js
+bring slack;
+bring cloud;
+
+let myBotToken = new cloud.Secret(name: "mybot_token");
+
+let slackbot = new slack.App(botToken: myBotToken);
+
+slackbot.onEvent("app_mention", inflight (ctx, event) => {
```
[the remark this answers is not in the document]
hasanaburayyan: Im not sure I understand what you mean both? There are many types of events "app_mention", "message", etc.. (https://api.slack.com/events)

slack/README.md
hasanaburayyan: No I dropped the websockets impl for sim since it made the switch from local development to AWS deployments more work in the slack portal. 

(requires disabling and enabling events API as you go back and forth)

slack/context.w
hasanaburayyan: Yea I can probably refactor this to make it not public.

I took the underscore convention from how the sim mocks in `bedrock` were defined.

slack/README.md
hasanaburayyan: Are you sure? Im pretty sure you were right the first time this is an `App` which can be used to describe a bot

slack/README.md
hasanaburayyan: Sure Ill add some docs for finding the id, I could have sworn it was only the id that was allowed but it actually looks like you can use a name. Ill update

slack/README.md
eladb: I am not sure :-)

So let's leave it `App`. Sorry!

slack/README.md
eladb: But then the app points only to one URL, no? So you can't both keep an app in production and in development.

Maybe the "correct" thing to do is to create two apps: one for dev and one for prod?

(we need to get this right or otherwise this won't hold water for real applications)

hasanaburayyan: > But then the app points only to one URL, no?

This is the case either way, when you enable socket mode slack will stop sending requests to your URL even if its specified (it confused the hell out of me why I was not getting messages for a while)

From slack:
```
Receive app payloads via Websockets instead of Request URLs
Turning on Socket Mode will route your app’s interactions and events over a WebSockets connection instead sending these payloads to Request URLs, which are public HTTP endpoints.

This setting is intended for internal apps that are in development or need to be deployed behind a firewall. It is not intended for widely distributed apps. Please set up Request URLs for your app before submitting to the App Directory. [Learn more](https://api.slack.com/socket-mode)
```

> Maybe the "correct" thing to do is to create two apps: one for dev and one for prod?

Yes it seems like the general guidance online is to use separate apps since there is not a concept of environments for slack Apps.

hasanaburayyan: Which maintaining two apps is kinda cumbersome, since if you have to update the manifest in dev, theres no easy copy this to another app button. Its a manual copy and paste job. Which I think I understand, but still sucks

PR #231, opened 2024-05-12
feat(python): support `cloud.Api`

eladcon:


python/lib.w
```diff
@@ -112,3 +114,30 @@ pub class InflightBucketEvent impl cloud.IBucketEventHandler {
     return this;
   }
 }
+
+pub class InflightApiOnRequest impl cloud.IApiEndpointHandler {
```
eladb, line 118:
```suggestion
pub class InflightApiEndpointHandler impl cloud.IApiEndpointHandler {
```

Or: `InflightApiRequestHandler`

python/lib.w
```diff
@@ -112,3 +114,30 @@ pub class InflightBucketEvent impl cloud.IBucketEventHandler {
...
+  _inflightType: str;
+  inner: types.IApiOnRequest;
+
+  new(props: types.InflightProps) {
+    this._inflightType = "_inflightPython";
+
+    let target = util.env("WING_TARGET");
```
eladb, line 125: @Chriscbr @hasanaburayyan what are we doing to get rid of this pattern?

Do we have a planned design? Do we have this in our tasks?

python/tfaws/api.js
```diff
@@ -0,0 +1,35 @@
...
+const { Function } = require("./function.js");
+
+module.exports.Api = class Api extends TfAwsApi {
+  constructor(
+    scope,
+    id,
+    props,
+  ) {
+    super(scope, id, props);
+  }
```
eladb, line 14:
```suggestion
```

I think this is not needed

python/tfaws/api.js
eladb, line 1: Can we change this to typescript?

python/tfaws/api_onrequest_inflight.w
```diff
@@ -0,0 +1,19 @@
+bring cloud;
+bring util;
+bring "../types.w" as types;
+bring "./inflight.w" as inflyght;
+
+pub class Inflight_tfaws extends inflyght.Inflight_tfaws impl types.IApiOnRequest {
```
eladb, line 6: @Chriscbr `internal` is direly missing

eladb, line 6: Why is there a difference between aws and sim?

python/wplatform.js
```diff
@@ -3,11 +3,13 @@ const { Function: TfAwsFunction } = require("./tfaws/function.js");
...
+const { Api: TfAwsApi } = require("./tfaws/api.js");
 
 const FUNCTION_FQN = "@winglang/sdk.cloud.Function";
 const QUEUE_FQN = "@winglang/sdk.cloud.Queue";
 const TOPIC_FQN = "@winglang/sdk.cloud.Topic";
 const BUCKET_FQN = "@winglang/sdk.cloud.Bucket";
+const API_FQN = "@winglang/sdk.cloud.Api";
 
 const createFunction = (target, scope, id, inflight, props) => {
   if (inflight._inflightType === "_inflightPython") {
```
eladb, line 15: Will this default to the normal function otherwise?

eladb (approved):

PR #232, opened 2024-05-14
fix(dynamodb): remove duplicate `Credentials` type

skyrpex:


Chriscbr (approved):

dynamodb/dynamodb.extern.d.ts
Chriscbr, line 1: Bump package.json

eladb: bumped

eladb (approved):

PR #240, opened 2024-05-19
feat(python): support dynamodb, sns, ses

eladcon:
Also:
* lambdas code now stored in s3
* support `cloud.Api` env vars.
* add local aws env vars when running in sim


python/sim/function.js
```diff
@@ -1,15 +1,16 @@
...
-    super(scope, id, inflight, props);
-    for (let e in props.env) {
-      inflight.inner.service.addEnvironment(e, props.env[e]);
+module.exports.handleSimInflight = (inflight, props) => {
+  for (let e in props.env) {
+    inflight.inner.service.addEnvironment(e, props.env[e]);
+  }
+
+  if (!App.of(inflight).isTestEnvironment) {
+    for (let key of ["AWS_REGION", "AWS_DEFAULT_REGION", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]) {
```
staycoolcall911, line 9: Does it throw an error if an env variable is missing?
Is it possible to throw an error in that case here?

python/test-assets/wing.py
```diff
@@ -219,3 +325,30 @@ def from_api_response(res = None):
     response["headers"] = res["headers"]
 
   return response
+
+class Aws:
```
staycoolcall911, line 329: Should this code be here?

python/sim/inflight.w
```diff
@@ -63,10 +59,43 @@ pub class Inflight impl cloud.IFunctionHandler {
...
     this.service = new cloud.Service(inflight () => {
       let clients = MutMap<Json>{};
-      for client in this.wingClients.entries() {
-        let value = MutJson this.clients.get(client.key);
-        value.set("handle", util.env(value.get("handle").asStr()));
-        clients.set(client.key, value);
+      for client in this.clients.entries() {
+        let value = client.value;
+
+        // TODO: move it to a function (there's a weird bug going on here)
```
staycoolcall911, line 65: Do you want to move this somewhere else?

staycoolcall911 (approved):
LGTM - left a few comments

PR #247, opened 2024-05-29
feat: some icons and colors

eladcon:


eladb (approved):
Swweeeeeeetn!

TIL that we can use icons

PR #250, opened 2024-05-30
feat: bring yaml

eladcon:


yaml/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+let file = fs.readFile("./test.yml");
+let value = yaml.parseValue(file);
+yaml.stringifyValue(json);
+
+new cloud.Function(inflight () => {
+  let value = yaml.parse(file);
+  yaml.stringify(json);
+});
+```
```
Chriscbr, line 30: We should totally include this in the built-in standard library!

eladb: Don't we already have some yaml stuff in the standard library?

eladb: I agree this should be part of the stdlib.

yaml/lib.w
```diff
@@ -0,0 +1,33 @@
+pub struct ParseOptions {
+  strict: bool?;
+}
+
+pub struct StringifyOptions {
+  indent: num?;
+}
+
+pub class Util {
```
eladb, line 9: what's the benefit of this duplication? Why not just make the extern methods public?

eladb (changes requested):

PR #257, opened 2024-06-06
fix(python): bucket test is broken on latest wing version

eladcon:


python/bucket.test.w
Chriscbr, line 1: Bump package.json

Chriscbr (approved):

PR #259, opened 2024-06-11
feat: bring momento

Chriscbr:
Code from episode #34 of the Wingly Update

Updated with a simulator implementation, and some better permission management.


momento/cache.w
```diff
@@ -0,0 +1,91 @@
+bring util;
+bring tf;
+
+pub struct CacheProps {
+  /// The name of the cache.
+  /// @default - a unique id will be generated
+  name: str?;
+  /// The default time-to-live for cache entries.
+  /// @default 60
```
eladb, line 9: Oh I wish we had default support in structs!

momento/cache.w
```diff
@@ -0,0 +1,91 @@
...
+    return Cache_tf._get(this.name, key);
+  }
+
+  pub inflight set(key: str, value: str, opts: CacheSetOptions?): void {
+    let ttl = opts?.ttl ?? 60s;
+    Cache_tf._set(this.name, key, value, ttl.seconds);
+  }
+
+  pub onLift(host: std.IInflightHost, ops: Array<str>) {
+    host.addEnvironment("MOMENTO_AUTH_TOKEN", util.env("MOMENTO_AUTH_TOKEN"));
```
eladb, line 72: Why not just use a cloud.Secret? It will behave the same way locally and will also work on the cloud in a secure way...

momento/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "url": "https://github.com/winglang/winglibs.git",
+    "directory": "momento"
+  },
+  "author": {
+    "email": "chrisr@wing.cloud",
+    "name": "Chris Rybicki"
+  },
+  "wing": {
+    "platforms": [
+      "sim",
```
eladb, line 16: Sim doesn't seem to be supported, no?

Chriscbr: Fixed

momento/cache.w
Chriscbr: I've reworked to use cloud.Secret now 👍 

It is a little weird in that the secret really isn't needed for the simulator but it seems to be OK

PR #261, opened 2024-06-13
chore(python): use `sim.Container`

eladcon:
Use the `sim.Container` to show containers logs. Also use docker image instead of local python to build the python dependencies.


staycoolcall911 (approved):
LGTM

PR #264, opened 2024-06-14
chore: add .gitattributes file

Chriscbr:


.mkrepo/gitattributes.w
```diff
@@ -0,0 +1,16 @@
+bring fs;
+
+pub class GitAttributes {
+  new() {
+    let lines = MutArray<str>[];
+    lines.push("# Auto-generated by ./mkrepo.sh");
+    lines.push("");
+    lines.push("/.github/workflows/*-pull.yaml linguist-generated");
```
eladb, line 8: I would have probably been nicer to invert the control for stuff like this (add a method `addGenerated(glob)` and call it from `Library`.

eladb (approved):

.mkrepo/gitattributes.w
Chriscbr: Yeah this was just quick and dirty 😅  it probably make sense to abstract it a bit more eventually

PR #266, opened 2024-06-17
chore: update winglibs to use `nodeof`

Chriscbr:
Updates libraries according to the breaking changes introduced in https://github.com/winglang/wing/pull/6721


staycoolcall911 (approved):
Looks good

PR #272, opened 2024-06-21
feat(dynamodb): add billingMode, deletionProtection props

thoroc:
Trying to address the missing properties coming from AWS CDK for DynamoDB Table.

Closes: [#6753](https://github.com/winglang/wing/issues/6753)


dynamodb/dynamodb-types.w
```diff
@@ -179,6 +192,9 @@ pub struct TableProps {
   timeToLiveAttribute: str?;
   globalSecondaryIndex: Array<GlobalSecondaryIndex>?;
   pointInTimeRecovery: bool?;
+  billingMode: BillingMode?;
+  removalPolicy: RemovalPolicy?;
+  deletionProtection: bool?;
```
Chriscbr, line 197: I'm thinking maybe we just keep `deletionProtection` for now since RemovalPolicy doesn't really have a Terraform equivalent AFAIK
```suggestion
  billingMode: BillingMode?;
  deletionProtection: bool?;
```

dynamodb/dynamodb.tf-aws.w
```diff
@@ -56,7 +61,12 @@ pub class Table_tfaws impl dynamodb_types.ITable {
       attribute: props.attributes,
       hashKey: props.hashKey,
       rangeKey: props.rangeKey,
-      billingMode: "PAY_PER_REQUEST",
+      billingMode: "{this.billingMode}",
+      removalPolicy: "{this.removalPolicy}",
```
Chriscbr, line 65: Looking from the build error I don't think this is a valid field https://github.com/winglang/winglibs/actions/runs/9635881681/job/26605621599?pr=272

dynamodb/dynamodb-types.w
thoroc: I have removed it as per suggestion. I still feel that needs to be there for anyone that would normally set those options in AWS CDK.

dynamodb/dynamodb.tf-aws.w
thoroc: Changed in c4d6624

dynamodb/dynamodb.tf-aws.w
```diff
@@ -56,7 +59,11 @@ pub class Table_tfaws impl dynamodb_types.ITable {
       attribute: props.attributes,
       hashKey: props.hashKey,
       rangeKey: props.rangeKey,
-      billingMode: "PAY_PER_REQUEST",
+      billingMode: "{this.billingMode}",
+      deletionProtectionEnabled: props.deletionProtection ?? false,
+      lifecycle: {
+        preventDestroy: props.deletionProtection ?? true
```
Chriscbr, line 65: Should these fields default to the same value?

dynamodb/tests/hasProperties.test.w
```diff
@@ -0,0 +1,36 @@
...
+
+  let tableDescription = simClient.describeTable({
+    TableName: tableWithBillingModeProvisioned.tableName,
+  });
+
+  expect.equal(
+    tableDescription.billingMode,
+    dynamodb_types.BillingMode.PROVISIONED
+  );
+}
```
Chriscbr, line 36: I think this test will fail if you try running it on the `tf-aws` platform since it's directly calling into the `dynamodb_sim` implementation. If we want to scope this test just to run on the simulator, you can wrap all of it in a block like:

```
bring util;

if util.env("WING_TARGET") == "sim" {
  // resources...
  // tests ...
}
```

Chriscbr (approved):
Looks great @thoroc! A few small comments but otherwise I'm cool merging it

dynamodb/dynamodb.tf-aws.w
thoroc: Oh that's a good catch. Yes that would make sense. I actually was wondering if there has been a wider discussion about having a system wide sensible default for those things.

Chriscbr: I don't think there's any built-in mechanisms at the moment -- perhaps we can add `deletionProtection` as a prop for more resources and eventually establish it as a convention of some sort.

thoroc:
Apologies, life got in the way of this PR. I'll get back to it asap.

PR #275, opened 2024-07-01
fix(cdk8s): unable to obtain root node

Chriscbr:
Fixes https://github.com/winglang/wing/issues/6806

### DIFF
--- a/.github/workflows/cdk8s-release.yaml
+++ b/.github/workflows/cdk8s-release.yaml
@@ -25,12 +25,7 @@ jobs:
         run: npm install --include=dev
         working-directory: cdk8s
       - name: Test
-        run: |
-          #!/bin/sh
-
-          wing compile -t ./lib/index.js examples/nodejs.main.w
-          wing compile -t ./lib/index.js examples/ubuntu.main.w
-          wing compile -t ./lib/index.js root.test.w
+        run: ./test.sh
         working-directory: cdk8s
       - name: Pack
         run: wing pack
--- a/.github/workflows/cdk8s-release.yaml
+++ b/.github/workflows/cdk8s-release.yaml
@@ -27,9 +27,10 @@ jobs:
       - name: Test
         run: |
           #!/bin/sh
-          cd examples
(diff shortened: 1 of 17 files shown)


.github/workflows/cdk8s-release.yaml
MarkMcCulloh, line 1: This file is marked as "generated" in the gitattributes, but it's only initially generated right?

cdk8s/examples/nodejs.main.w
```diff
@@ -1,8 +1,9 @@
 bring "cdk8s-plus-27" as k8s;
+bring fs;
 
 // lets create a volume that contains our app.
 let appData = new k8s.ConfigMap();
-appData.addDirectory("./nodejs-app");
+appData.addDirectory(fs.join(@dirname, "./nodejs-app"));
```
MarkMcCulloh, line 6: nit:

```suggestion
appData.addDirectory(fs.join(@dirname, "nodejs-app"));
```

cdk8s/root.test.w
MarkMcCulloh, line 1: out of context this test seems strange imo, maybe it's worth linking to the issue

.github/workflows/cdk8s-release.yaml
```diff
@@ -27,9 +27,10 @@ jobs:
       - name: Test
         run: |
           #!/bin/sh
-          cd examples
-          wing compile -t ../lib/index.js nodejs.main.w
-          wing compile -t ../lib/index.js ubuntu.main.w
+
+          wing compile -t ./lib/index.js examples/nodejs.main.w
```
MarkMcCulloh, line 31: We could run the test.sh here instead, to avoid out-of-sync testing scenarios

MarkMcCulloh (approved):

.github/workflows/cdk8s-release.yaml
Chriscbr: Sure, let me try that

Chriscbr: Done

PR #276, opened 2024-07-01
chore(dynamodb): make test.sh executable

MarkMcCulloh:


Chriscbr (approved):

dynamodb/test.sh
Chriscbr, line 1: The file's empty? Do we need it?

Chriscbr: If it's an intentional no-op maybe we can add a `# skip tests` comment or something

MarkMcCulloh: It's not empty, although I guess the gh UI isn't showing anything because there was no content change https://github.com/winglang/winglibs/blob/mark/make-dynamo-db-x/dynamodb/test.sh

PR #279, opened 2024-07-03
feat(stripe): added new winglib for stripe

boyney123:
Here is my first attempt at a winglib 🎆

### Core features

- This wing lib allows developers to quickly create stripe webhooks and subscribe to any stripe event.
- Creating a Stripe webhook will create a API, Queue and Function cloud resources.
- README has more information on how to use it.

### Questions

- Would love feedback on the pattern, is this what you expect from a winglib? Does it make sense? Could things be done better?
- I struggled to test this. I wanted to mock out the Stripe webhook verification code (extern), I'm passing environment variable to my `/webhook` endpoint to  make it optional, but not sure if this is best or not?
- I want to change my ENV variable value `IGNORE_STRIPE_WEBHOOK_VALIDATION` in my wing test, is this possible? I want a test to set it to `false` so I can test my edge cases..?
- @Chriscbr helped me understand why my ENV variables in my extern (inflight) were not present, I expected my preflight ENV variables to be accessible here.... he mentioned @eladb you spoke about this before? Shall we raise an issue for this? Or is this something that doesnt make sense? I'm still wrapping my head around some of this so more than likely just me being a newb here too!
- I wanted an `.env.test` file to set my `IGNORE_STRIPE_WEBHOOK_VALIDATION` to true. This file did not get pick up by my tests when running `wing test`, is this a bug? From what I read in the docs this is expected? Unless I misread!

Would love some thoughts/advice, I created this to gain more understanding of wing, whilst also hopefully providing value to others with a new lib :)


stripe/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Example
+
+```js
+bring cloud;
+bring "./lib/stripe.w" as Stripe;
+
+let stripeSecretKey = new cloud.Secret(name: "STRIPE_API_TOKEN") as "Stripe Secret";
+let webhookSecret = new cloud.Secret(name: "STRIPE_ENDPOINT_SECRET") as "Webhook Secret";
+let stripe = new Stripe.Webhook({ webhookSecret: webhookSecret, secretKey: stripeSecretKey }) as "Stripe integration";
```
eladb, line 29:
```suggestion
let stripe = new Stripe.Webhook(webhookSecret: webhookSecret, secretKey: stripeSecretKey) as "Stripe integration";
```

Chriscbr:
> I want to change my ENV variable value IGNORE_STRIPE_WEBHOOK_VALIDATION in my wing test, is this possible? I want a test to set it to false so I can test my edge cases..?

IIRC we have an API that lets you check if you're running in a wing test or not -- so you might be able to add some logic to set an environment variable differently by adding a conditional:

```js
bring cloud;

let isTest = nodeof(this).app.isTestEnvironment; // bool

new cloud.Function(inflight () => {
  log(isTest); // "true" if this function is invoked in a wing test, "false" otherwise
});
```

Something about this API feels clunky though, @eladb wondering if you have ideas for renaming or a better place to put this?

stripe/lib.w
```diff
@@ -0,0 +1,126 @@
+bring util;
+bring cloud;
+bring "./types.w" as types;
+
+/// Properties for Stripe webhook
+pub struct Props {
+  /// The secret key for Stripe as a string
+  secretKey: str?;
+  /// Secret key for Stripe as a Secret resource (recommended)
+  secretKeySecret: cloud.Secret?;
```
eladb, line 10: Why do we need both options?

eladb:
Clunky is the understatement of the year...

Maybe we can sugar `nodeof(this).app` as `@app`?

stripe/lib.w
boyney123: Looked at the OpenAPI example, I did mainly for testing to be honest. Being able to create a Stripe client without secrets defined just passing a string in, seems odd though.

I need to explore how tests mock the Secret resource....and understand this more.

boyney123:
> > I want to change my ENV variable value IGNORE_STRIPE_WEBHOOK_VALIDATION in my wing test, is this possible? I want a test to set it to false so I can test my edge cases..?
> 
> IIRC we have an API that lets you check if you're running in a wing test or not -- so you might be able to add some logic to set an environment variable differently by adding a conditional:
> 
> ```js
> bring cloud;
> 
> let isTest = nodeof(this).app.isTestEnvironment; // bool
> 
> new cloud.Function(inflight () => {
>   log(isTest); // "true" if this function is invoked in a wing test, "false" otherwise
> });
> ```
> 
> Something about this API feels clunky though, @eladb wondering if you have ideas for renaming or a better place to put this?

Thanks will have a look. 

FWIW it does initially feel odd to me to change implementation code to add test cases in there, rather than somehow mock this stuff in the test itself? Does that make sense? Not sure if it's cos I'm coming from a TS/JS background this is what I expect, or if this is normally fine and to be expected?

Do we expect our users to change libs/implementation code with test edge cases?

Chriscbr:
> Clunky is the understatement of the year...
> 
> Maybe we can sugar `nodeof(this).app` as `@app`?

I like it!

> FWIW it does initially feel odd to me to change implementation code to add test cases in there, rather than somehow mock this stuff in the test itself? Does that make sense? Not sure if it's cos I'm coming from a TS/JS background this is what I expect, or if this is normally fine and to be expected?
> 
> Do we expect our users to change libs/implementation code with test edge cases?

Mmm - I think I get what you mean, coupling the logic closer to the test definitions sounds like it might be a better place.

I think the behavior you were expecting (where .env.test variables are automatically loaded) probably should be supported. Smells like a bug - could you open an issue for it?

Perhaps we can try going forward with the changes proposed in https://github.com/winglang/wing/issues/4601, so all environment variables are retrieved during preflight. Then if you want to customize the value of "IGNORE_STRIPE_WEBHOOK_VALIDATION", I think you should be customize it through either the `.env.test` file or by setting it explicitly:

```js
bring cloud;
bring util;

// mylib.w

class Foo {
  f: cloud.Function;
  new() {
    let key = util.tryEnv("IGNORE_VALIDATION");
    this.f = new cloud.Function(inflight () => {
      log(key ?? "nil");
    });
  }
  pub inflight run() {
    this.f.invoke();
  }
}

// test.w

util.setEnv("IGNORE_VALIDATION", "true");
let f = new Foo();

test "my test" {
  f.run(); // logs "true"
}
```

PR #286, opened 2024-07-10
fix: mkrepo main for wingcloud deployment

polamoros:


.mkrepo/main.w
```diff
@@ -29,7 +29,11 @@ for file in fs.readdir(".") {
   libs.push(lib);
 }
 
-readme.update(libs.copy());
+try {
+  readme.update(libs.copy());
```
eladb, line 33: Why is this needed?

eladb (changes requested):

.mkrepo/main.w
polamoros: I was trying it out cause I'm getting this error on wingcloud

image

https://wing.cloud/polamoros/winglibs/logs/main

PR #287, opened 2024-07-11
fix: dynamodb optional json use

tsuf239:
After merging https://github.com/winglang/wing/pull/6839
everything is passing:
image


tsuf239:
https://github.com/winglang/winglibs/blob/48044f0a1f00e03218691dc017eecf513078a902/dynamodb/package.json#L3
@eladcon here?

PR #288, opened 2024-07-14
fix(dynamodb): adding global index to tf-aws target

marciocadev:
The `compatibility.gsi.test.w` test was not working on AWS, so I added the configuration during table creation and the permission to access the global index.


marciocadev:
> Confirmed that the test passed on tf-aws:
> image
> (and those that didn't pass failed on unrelated naming issues)

Yes, I open another PR to fix this too @tsuf239

PR #290, opened 2024-07-15
fix(dynamodb): adding the update policy

marciocadev:


eladb (approved):
Do we have a test for this somehow?

PR #293, opened 2024-07-18
chore: automated version bump

marciocadev:
I always forgot to update the version in package.json when changing some library, and if there was more than one update in the same library, this could also cause versioning issues.

With this update, the version in package.json will be updated before packing, and versioning issues will no longer occur.


.github/workflows/bedrock-release.yaml
```diff
@@ -27,6 +27,16 @@ jobs:
...
+      - name: Automated version bump
+        uses: phips28/gh-action-bump-version@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.PROJEN_GITHUB_TOKEN }}
+          PACKAGEJSON_DIR: bedrock
+        with:
+          minor-wording: add,Adds,new
+          major-wording: MAJOR,cut-major
+          patch-wording: patch,fixes
+          rc-wording: RELEASE,alpha
```
Chriscbr, line 39: I'm not sure this approach will work for us.
1. It will push a new commit back to the repo with the version bump -- but this will cause the the `lib-release.yaml` workflow to run again (causing a feedback loop).
2. Since the action is run once for each package, one PR that changes multiple libraries may result in multiple version bumps, which will pollute the commit history quite a bit.
3. To follow semver for pre-1.0 libraries and be consistent with the `winglang` repo, we only want breaking changes to result in a bump to the minor version (e.g. 0.1.0 to 0.2.0) -- all other changes should bump the patch version (e.g. 0.1.0 to 0.1.1)

Perhaps we can start by adding a script / check that runs over the repo in CI, compares the files on the current branch to the files on the main branch, and requires that if files in the `libname` folder are changed, then the "version" must be changed in `libname/package.json`. Maybe we could have it automatically post a GitHub comment as a reminder as well.

Chriscbr (changes requested):

.github/workflows/bedrock-release.yaml
marciocadev: You are right, I did several tests and by changing to this

```yaml
- name: Automated version bump
  uses: phips28/gh-action-bump-version@master
  env:
    GITHUB_TOKEN: ${{ secrets.PROJEN_GITHUB_TOKEN }}
    PACKAGEJSON_DIR: bedrock
  with:
    minor-wording: "BREAKING CHANGE"
    major-wording: ""
    patch-wording: "feat,fix,chore"
    rc-wording: ""
```

in theory, it should work. Although it works with the patch, the plugin is not able to capture the BREAKING CHANGE.

I am marking this PR as a draft and studying how to proceed. In the worst case, I will try to create a GitHub action to solve this.

marciocadev: > 2. Since the action is run once for each package, one PR that changes multiple libraries may result in multiple version bumps, which will pollute the commit history quite a bit.

You are right on this point, Chris, but a PR that changes more than one library should really generate new versions of the altered libraries, don't you think?

Chriscbr: > You are right on this point, Chris, but a PR that changes more than one library should really generate new versions of the altered libraries, don't you think?

Definitely - perhaps there's a way so the version bumps can all be done in a single commit?

PR #301, opened 2024-07-30
feat: cloudv2

Chriscbr:
Begin the process of migrating built-in `cloud` resource implementations from TypeScript to Wing. Only `sim` and `tf-aws` targets are in scope at the moment.

- [x] Counter
- [ ] ... the rest

I put this into a new winglib just to avoid the initial issue of stuffing the source code into the bundled `winglang` package and wiring the compiler to auto-include those source files -- but eventually the code will probably go there.

The implementation of `Counter` is almost working, but I believe there's a lifting issue causing the tests to fail on AWS (https://github.com/winglang/wing/issues/6960) - I need to look into this.

### DIFF
--- a/cloudv2/counter.w
+++ b/cloudv2/counter.w
@@ -1,0 +1,235 @@
+bring aws;
+bring cloud;
+bring fs;
+bring ui;
+bring util;
+bring sim;
+
+bring "@cdktf/provider-aws" as tfaws;
+bring "./util.w" as myutil;
+
+// TODO: Default values for struct fields - https://github.com/winglang/wing/issues/3121
+
+pub struct CounterProps {
+  /// The initial value of the counter
+  /// @default 0
+  initial: num?;
+}
+
+interface ICounter {
+  /// Increments the counter atomically by a certain amount and returns the previous value.
+  /// - `amount` The amount to increment by (defaults to 1)
+  /// - `key` The key of the counter (defaults to "default")
+  inflight inc(amount: num?, key: str?): num;
+
+  /// Decrements the counter atomically by a certain amount and returns the previous value.
+  /// - `amount` The amount to decrement by (defaults to 1)
+  /// - `key` The key of the counter (defaults to "default")
+  inflight dec(amount: num?, key: str?): num;
+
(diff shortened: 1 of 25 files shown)


cloudv2/test/counter/dec.test.w
```diff
@@ -0,0 +1,47 @@
+bring "../../" as cloud;
```
eladb, line 1: are these basically the tests from `sdk_tests`?

Let's make sure to delete them once we've migrated.

cloudv2/counter.w
```diff
@@ -0,0 +1,237 @@
...
+  pub peek(key: str): num {
+    return this.values.tryGet(key) ?? this.initial;
+  }
+
+  pub set(value: num, key: str) {
+    this.values[key] = value;
+  }
+}
+
+class Counter_sim impl ICounter {
```
eladb, line 132: I would put each implementation in a separate file

cloudv2/counter.w
```diff
@@ -0,0 +1,237 @@
...
+    let target = util.env("WING_TARGET");
+    let id = nodeof(this).id;
+    this.initial = props.initial ?? 0;
+    if target == "sim" {
+      this.inner = new Counter_sim(props) as id;
+    } elif target == "tf-aws" {
+      this.inner = new Counter_tfaws(props) as id;
+    } else {
+      throw "Unsupported target: " + target;
+    }
```
eladb, line 58: This means that I cannot override the implementation with a platform extension, correct?

I suspect we need to address this before we finish the migration because we have users who rely on this capability.

eladb (approved):

cloudv2/counter.w
Chriscbr: Yap

cloudv2/test/counter/dec.test.w
Chriscbr: Yep, pretty much copied fully

cloudv2/counter.w
Chriscbr: Done

PR #311, opened 2024-08-06
fix(cloudv2)!: hide internal APIs

Chriscbr:


cloudv2/counter-aws.w
```diff
@@ -9,7 +9,7 @@ pub interface IAwsCounter {
   dynamoTableName(): str;
 }
 
-pub class Counter_tfaws impl counter_types.ICounter, IAwsCounter {
+internal class Counter_tfaws impl counter_types.ICounter, IAwsCounter {
```
eladb, line 12: Very cool with `internal` but I was actually wondering if it might be useful to expose these classes in case the user wants to explicitly instantiate an AWS counter (maybe under the `aws` namespace):

```js
bring cloud;

new cloud.aws.Counter();
```

Chriscbr: Yeah, I think that's probably a good call

PR #313, opened 2024-08-30
chore: add docs to libraries

Chriscbr:
Using the updated scripts `./gen-docs-all.sh` to generate docs in all winglibs, and `./bump-all.sh` to bump the versions in package.json. Any libraries that are broken are skipped for now.

### DIFF
--- a/.mkrepo/main.w
+++ b/.mkrepo/main.w
@@ -9,6 +9,7 @@ bring "./stale.w" as stale;
 bring "./readme.w" as readme;
 
 let workflowdir = ".github/workflows";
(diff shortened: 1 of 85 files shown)


Chriscbr:
Getting some weird build error where running `wing gen-docs` in the `k8s` folder is hanging endlessly. Not sure what the cause is. https://github.com/winglang/winglibs/actions/runs/10636626654/job/29488996612?pr=313

Going to see if we can ignore it for now...

eladb (commented):
Looks awesome

boyney123:
Nice work @Chriscbr , will automate this into our docs now :)

PR #315, opened 2024-08-30
WIP: testing

Chriscbr:
Trying to root cause why `wing gen-docs` is hanging for `@winglibs/k8s` inside CI/CD


Chriscbr:
Through some imprecise debugging, it seems like I was able to narrow down that the issue of hanging comes up whenever there is a `lib/` directory inside of the project. The issue seems to be specific to the GitHub runners (I'm unable to reproduce it locally) and I'm also unable to figure out why it only happens with the `wing gen-docs` command.

From the logging I was able to add, it seems like the hanging is happening while "preopening" files for the WebAssembly module. https://github.com/winglang/wing/blob/b9764935a94c38a561e6c9ac5f454976f2d86d7d/libs/wingcompiler/src/wingc.ts#L82	 Perhaps it's trying to "preopen" all files from the file system?

I have an idea for a small patch that might update the relevant behavior so the CLI doesn't try opening as many files. Alas, I'm not sure if it will fix the issue, but we can try and see.

Chriscbr:
I'd categorize this as a type of bug that likely could have been avoided if we didn't have to use WebAssembly as an intermediary between the CLI and compiler.

Chriscbr:
The issue has been fixed with #7061 - reran the release workflow here (https://github.com/winglang/winglibs/actions/runs/10636846595/job/29498287120) and the new version of k8s with docs has been published.

PR #322, opened 2024-09-22
feat(dynamodb): add `batchGet` and `batchWrite` methods

co-go:
Adding in [`BatchWriteItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) and [`BatchGetItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html) methods to the DynamoDB library. Normally these methods operate at the client level (as opposed to the "table" level like the rest of the library) due to the fact that they _can_ be called across multiple tables in a single API call. Given that most folks will likely just want to be able to call the batch methods on the single Dynamo tables they have created, I added the batch functions to the `ITable` as well as the `IClient` so the caller has flexibility.

### DIFF
--- a/dynamodb/dynamodb-client.w
+++ b/dynamodb/dynamodb-client.w
@@ -64,7 +64,7 @@ pub inflight class Client impl dynamodb_types.IClient {
   pub inflight update(options: dynamodb_types.UpdateOptions):  dynamodb_types.UpdateOutput {
     let input: MutJson = options;
     input.set("TableName", this.tableName);
-    return unsafeCast(this.client.update(input)); 
+    return unsafeCast(this.client.update(input));
   }
 
   pub inflight transactWrite(options: dynamodb_types.TransactWriteOptions): dynamodb_types.TransactWriteOutput {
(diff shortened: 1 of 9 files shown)


dynamodb/dynamodb-client.w
```diff
@@ -114,4 +114,32 @@ pub inflight class Client impl dynamodb_types.IClient {
...
+
+      for PutRequest in item.value.PutRequests ?? [] {
+        tableReqs.push({ PutRequest });
+      }
+
+      reqItems.set(item.key, tableReqs);
+    }
+
+    return unsafeCast(this.client.batchWrite({
+      RequestItems: Json.parse(Json.stringify(reqItems)),
```
co-go, line 140: Sorry, new to Wing! There is probably a better way to get this thing into a Json type, I just couldn't seem to figure it out.

dynamodb/package-lock.json
```diff
@@ -17,413 +17,4774 @@
         "aws-sdk": "^2.1610.0"
       }
     },
-    "node_modules/@aws-crypto/crc32": {
+    "node_modules/@aws-crypto/sha256-browser": {
```
co-go, line 20: I ran an `npm audit fix` which seemed to have made most of these tweaks. Let me know if I should revert this!

eladb:
Awesome! We will review and get back to you shortly!

co-go:
Forgot docs and IAM access needed. Should be ready for review now!

dynamodb/dynamodb-types.w
```diff
@@ -153,6 +153,94 @@ pub struct TransactWriteOptions {
...
+  /// data types must match those of the table's key schema. If any attributes are present in the item
+  /// that are part of an index key schema for the table, their types must match the index key
+  /// schema.
+  Item: Json;
+}
+
+/// Represents an operation to perform - either `DeleteItem` or `PutItem`. Used internally to map
+/// from the `DeleteRequests` or `PutRequests` on `TableBatchWriteOptions` to the contract required
+/// from DynamoDB.
+pub struct WriteRequest {
```
Chriscbr, line 204: It looks like this struct isn't used publicly anywhere (like in as a parameter type or return type on the batchWrite method), so what we can do is make it `internal` instead so it's hidden from users importing the library.

The motivation here is to reduce extra API surface area when people are going through docs etc.
```suggestion
internal struct WriteRequest {
```

dynamodb/dynamodb-client.w
Chriscbr: One way to get mutable collections in Wing into Json is by obtaining immutable copies with the `.copy()` method, and then Wing lets you use many immutable arrays or collections in place of Json. But here it looks like the collection is "deep" so that might not work here. I think if we added some `toJson()` and `toMutJson()` methods to the language, perhaps that would be more ideal. I'll take a note and open a separate issue - thanks for the heads up.

Chriscbr (changes requested):
This is an excellent addition @co-go. I've requested one minor change, otherwise it looks great 👍

dynamodb/dynamodb-client.w
co-go: Good to know– that would be a useful set of functions to have!

Chriscbr (approved):

PR #330, opened 2025-01-19
initial work on pulumi platform provider

eladb:


pulumi/README.md
```diff
@@ -0,0 +1,27 @@
...
+```sh
+npm i @winglibs/pulumi
+```
+
+## Usage
+
+```js
+bring pulumi;
+
+let adder = new pulumi.Adder();
```
ShaiBer, line 18: I see you also added support for compiling some standard lib resources to pulumi (bucket, queue and function). Should this be documented here too?

